PR #366, opened 2026-06-11
fix(dhcp): #365 group socket-mode (raw/udp) so directly-attached clients get leases

### LINKED ISSUE #365: DHCP: Kea rendered with dhcp-socket-type "udp" — directly-attached clients get no lease (no operator control, wrong default)
## Summary

The DHCP agent always renders Kea with **`dhcp-socket-type: udp`**, and there is **no operator control** to change it. UDP datagram sockets can't reliably receive broadcast `DHCPDISCOVER`s from **directly-attached (non-relayed)** clients, so on the all-in-one appliance — or any subnet where Kea is on the same L2 as its clients — Kea is silently deaf: DISCOVER is on the wire, the firewall is open, the Kea process is healthy, but **no `DHCPOFFER` is ever sent**.

## Symptom (field-reproduced)

A client on the same vSwitch / L2 as the appliance gets no lease:

- Wireshark on the client + `tcpdump -ni any 'port 67 or 68'` on the appliance both show the broadcast `DHCPDISCOVER` arriving on the host NIC.
- Firewall is open — `nft list chain inet filter input` shows `udp dport 67 accept` / `udp dport 68 accept`.
- The Kea pod is `healthy`; its log shows heartbeats + `statistic-get-all` but **never logs receiving the DISCOVER** and never emits an OFFER.
- Static IP on the same client works fine (L2/L3 connectivity is not the problem).

## Root cause

- `agent/dhcp/spatium_dhcp_agent/render_kea.py:545` renders `"dhcp-socket-type": server.get("dhcp_socket_type", "udp")`.
- The control plane **never sends a `server` block** in the config bundle (`backend/app/services/dhcp/config_bundle.py` has no such key), and there is **no UI/API field** for socket type anywhere in `backend/` or `frontend/`. So the rendered value is *always* `udp`.
- `udp` sockets are intended for **relayed** traffic. Kea's own default is `raw` (AF_PACKET), which receives broadcasts from clients that don't have an IP yet. **`raw` is the superset** — it serves *both* direct and relayed clients; `udp` is relay-only.
- The appliance DaemonSet already grants `NET_RAW` (`charts/spatiumddi-appliance/templates/dhcp-kea.yaml:66`) — the exact capability raw sockets need — but the rendered config never uses it.

## Why this can't be a per-subnet toggle

`dhcp-socket-type` lives in Kea's `interfaces-config`, which is **per-daemon** (one value per `kea-dhcp4` process). It cannot vary per subnet/scope. The control must therefore live at **server-group level** (consistent across HA peers in the group).

## Proposed fix

- **Model:** add `DHCPServerGroup.dhcp_socket_mode` — `direct` → `raw` (**default**), `relay` → `udp`. Migration backfills existing rows to `direct`.
- **Bundle:** emit it in the DHCP config bundle (`server` block) and fold it into the bundle ETag so a change propagates over the long-poll; `collect_wake` the affected DHCP channel after commit.
- **Render paths:** honor it in both the agent (`render_kea.py`) and the agentless driver (`backend/app/drivers/dhcp/kea.py` `interfaces-config`); change the agent's fallback default from `udp` → `raw`.
- **Capability:** add `NET_RAW` to the docker-compose Kea services (`docker-compose.yml`, `docker-compose.dev.yml`, `docker-compose.agent-dhcp.yml`) so `raw` works out of the box (the appliance already has it). Operators who can't grant the cap pick **Relay-only**.
- **Frontend:** a "Client reachability" toggle on the DHCP group create/edit modal — *Directly attached / mixed (raw sockets, recommended)* vs *Relay-only (UDP sockets)*, with help text noting raw also handles relayed traffic.
- **MCP:** carry the field on the DHCP-group `propose_*` tool(s) where present.
- **Docs:** `docs/features/DHCP.md`, `docs/drivers/DHCP_DRIVERS.md`, and the `docs/TROUBLESHOOTING.md` matrix ("logs nothing + firewall open → it's the socket type, not the firewall").

## Acceptance criteria

- A fresh appliance with a directly-attached subnet leases to a broadcast client with **no manual steps**.
- Toggling a group to **Relay-only** renders `udp`; back to **Directly attached** renders `raw`; agents pick up the change within a heartbeat (ETag shift).
- Existing installs default to `raw` on upgrade; docker-compose Kea services carry `NET_RAW`.

mzac:
Closes #365.

## Problem

The DHCP agent always rendered Kea with **`dhcp-socket-type: udp`**, with no operator control. UDP datagram sockets are relay-only — they can't receive broadcast `DHCPDISCOVER`s from directly-attached clients that have no IP yet. So on the all-in-one appliance (or any on-LAN DHCP server), Kea was alive and healthy but **deaf to the broadcast**: DISCOVER on the wire, firewall open, no `DHCPOFFER`. Field-reproduced; the smoking gun is `render_kea.py` defaulting to `udp` while the control plane never sent a socket-type and there was no UI/API knob.

## Fix

A per-**group** knob `dhcp_socket_mode` — **`direct` → `raw`** (default) / **`relay` → `udp`**. `dhcp-socket-type` is a per-daemon Kea setting (can't vary per subnet), so it lives on `DHCPServerGroup` and applies to every member Kea. `raw` (AF_PACKET) hears broadcast DISCOVERs **and** relayed traffic — the superset, and Kea's own default; `relay`/`udp` is the restricted opt-out for relay-only or capability-restricted runtimes.

It's delivered to every deploy path (k8s / compose / appliance) through the **ConfigBundle long-poll**, folded into the bundle **ETag** so a change re-renders the running Kea within a heartbeat — not via the supervisor role assignment, since the agent renders `interfaces-config` from the bundle. Distinct from the existing `network_mode` (host/bridged container namespace).

## What changed

- **Model + migration** `f3b9c1d6a274` — backfills existing groups to `direct` (so upgrades start answering directly-attached clients with no action).
- **Bundle** — `config_bundle` maps mode → socket type; new `ConfigBundle.dhcp_socket_type` in `compute_etag()`; agent wire gains a `server` block.
- **Renderers** — agentless `kea.py` Dhcp4 honors it (v6 excluded — DHCPv6 has no socket-type); agent `render_kea.py` fallback flipped `udp` → `raw`.
- **API** — group `Create/Update/Response` + validator + wake-on-change.
- **Compose** — `NET_RAW` granted on all Kea services in `docker-compose{,.dev,.agent-dhcp}.yml` (the appliance DaemonSet already had it); corrected the stale "No NET_RAW" comments.
- **Frontend** — api types + a "Client reachability" toggle on the group modal.
- **MCP** — `list_dhcp_server_groups` surfaces `dhcp_socket_mode`.
- **Docs** — DHCP.md, DHCP_DRIVERS.md, and the TROUBLESHOOTING.md "no DHCPOFFER" matrix.

## Verification

- `ruff` / `black` / `mypy` (backend) clean; frontend `tsc` clean.
- New `test_dhcp_socket_mode.py` — **9/9 pass** (driver render raw/udp, default-raw, v6-excluded, ETag shift, group→bundle mapping, API round-trip + validation).
- Migration applies cleanly (`b6f4d2a91c83 → f3b9c1d6a274`, column default `direct`).
- **Live on a dev stack**: recreated `dhcp-kea` with `NET_RAW`; the agent re-rendered `"dhcp-socket-type": "raw"` and Kea came up running on raw sockets.

### DIFF
--- a/agent/dhcp/images/kea/Dockerfile
+++ b/agent/dhcp/images/kea/Dockerfile
@@ -41,6 +41,11 @@ FROM alpine:3.22 AS runtime
 # re-resolved against the current alpine 3.22 index. GitHub Actions
 # Docker layer cache (``cache-from: type=gha``) is otherwise happy
 # to serve a 30-day-old ``apk add`` layer that Trivy will then flag.
+#
+# ``libcrypto3``/``libssl3`` pinned ``>=3.5.7-r0`` for CVE-2026-45447
+# (HIGH, OpenSSL — alpine 3.22 base ships 3.5.6-r0, fixed in 3.5.7-r0):
+# the explicit lower bound installs the patched openssl libs AND busts
+# the ``cache-from`` layer so a stale apk add can't reintroduce 3.5.6-r0.
 RUN apk upgrade --no-cache \
  && apk add --no-cache \
       'kea>=2.6.5-r0' kea-dhcp4 kea-dhcp6 kea-ctrl-agent kea-admin \
@@ -49,6 +54,7 @@ RUN apk upgrade --no-cache \
       python3 py3-pip libpcap \
       libcap \
       ca-certificates tzdata \
+      'libcrypto3>=3.5.7-r0' 'libssl3>=3.5.7-r0' \
  && setcap cap_net_bind_service=+ep /usr/sbin/kea-dhcp4 \
  && setcap cap_net_bind_service=+ep /usr/sbin/kea-dhcp6 \
  && addgroup -S spatium \
--- a/agent/dhcp/spatium_dhcp_agent/render_kea.py
+++ b/agent/dhcp/spatium_dhcp_agent/render_kea.py
@@ -542,7 +542,14 @@ def render(
     dhcp4: dict[str, Any] = {
         "interfaces-config": {
             "interfaces": list(interfaces),
-            "dhcp-socket-type": server.get("dhcp_socket_type", "udp"),
+            # Issue #365 — default to ``raw`` (AF_PACKET) so Kea hears
+            # broadcast DISCOVERs from directly-attached clients. The
+            # control plane sends the resolved value (``raw`` for group
+            # socket_mode "direct", ``udp`` for "relay"); the ``raw``
+            # fallback only applies to bundles from an older control plane
+            # that predates the ``server`` block. ``raw`` needs CAP_NET_RAW
+            # (granted on the appliance DaemonSet + compose Kea services).
+            "dhcp-socket-type": server.get("dhcp_socket_type", "raw"),
         },
         "control-socket": {
             "socket-type": "unix",
--- a/agent/dns/images/bind9/Dockerfile
+++ b/agent/dns/images/bind9/Dockerfile
@@ -26,6 +26,11 @@ FROM alpine:3.22 AS runtime
 # ``cache-from: type=gha`` on the build workflow) is otherwise
 # perfectly happy to serve a 30-day-old ``apk add`` layer that
 # Trivy will then flag.
+#
+# ``libcrypto3``/``libssl3`` pinned ``>=3.5.7-r0`` for CVE-2026-45447
+# (HIGH, OpenSSL — alpine 3.22 base ships 3.5.6-r0, fixed in 3.5.7-r0):
+# the explicit lower bound installs the patched openssl libs AND busts
+# the ``cache-from`` layer so a stale apk add can't reintroduce 3.5.6-r0.
 RUN apk upgrade --no-cache \
  && apk add --no-cache \
       bind bind-tools \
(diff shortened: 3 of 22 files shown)


mzac:
### Note: bundling an OpenSSL package upgrade in this PR

The `build (kea)` Trivy scan surfaced **CVE-2026-45447** (HIGH) in `libcrypto3`/`libssl3` — the OpenSSL libs in the `alpine:3.22` base image (`3.5.6-r0`, fixed in `3.5.7-r0`). It's an **upstream base-image CVE unrelated to the DHCP socket-mode change** (this branch only touches Python/TS/compose); `main`'s kea build was green on 2026-05-31 and the CVE was disclosed since. The `apk upgrade --no-cache` step couldn't pull the fix because the BuildKit `cache-from: type=gha` layer was stale from before the patch shipped.

Rather than leave it for a follow-up, I've **upgraded the packages here at the same time** — pinning `libcrypto3>=3.5.7-r0` + `libssl3>=3.5.7-r0` in the `apk add` line of **all three agent images** (kea, bind9, powerdns), the same explicit-lower-bound trick these Dockerfiles already use (e.g. `kea>=2.6.5-r0`) to both install the patched version and bust the stale cache layer. Verified locally that the pin resolves to `3.5.7-r0`.

So this PR now carries two things: the #365 socket-mode fix, and the openssl bump that clears the Trivy finding.

Issue #312, opened 2026-05-29
Installer: role-aware sizing warning for Control plane (24 GiB floor installs but can't run it → DiskPressure on first boot)


mzac:
## Problem

The `spatium-install` wizard enforces a single hard floor for **all**
roles — **24 GiB disk / 2 GiB RAM** — below which it refuses to
install. That floor only guarantees the appliance can *install and
boot*. It is **not** enough to *run the Control plane role*, which is
the heaviest one (k3s etcd + CloudNativePG Postgres + Redis +
api/worker/beat/frontend, plus the HA manifests that pull 2× replicas).

A field tester provisioned a VMware Workstation VM at roughly the floor
(~24 GiB disk, 3.8 GiB RAM) in **Control plane + frontend** mode. It
installed cleanly, then died on first boot in a **DiskPressure eviction
storm**:

- The 24 GiB disk gives `/var` only ~7 GiB (`var = disk − ~17 GiB` after
  the ESP + STATE + 2× 8 GiB A/B slots).
- First boot imports/pulls the full control-plane image set into
  containerd on `/var`, *and* the local-path PVCs for Postgres + Redis
  also land on `/var`.
- `/var` crossed the kubelet's default nodefs eviction threshold
  (~10% free). The eviction manager started killing pods
  (`postgresql-initdb` Failed, `spatium-supervisor` Evicted) and image
  GC began deleting images **mid-pull** — a thrash loop that never
  converges.

Console symptoms (from the tester's screenshot): `/var 65%` and
climbing; live log full of *"Eviction manager: attempting to reclaim
ephemeral-storage"*, *"must evict pod(s)"*, *"Removing image to free
bytes"*; `cnpg postgresql-initdb` **Failed**, `spatium-supervisor`
**Evicted/Failed**.

The recommended sizing is already documented (**Control plane =
4 vCPU / 8 GiB / 40 GiB SSD**, README + APPLIANCE.md §"Recommended
sizing") — but nothing in the installer surfaces it at the moment the
operator picks the Control plane role, so the floor reads as
"24 GiB is fine."

## Proposal

Make the installer role-aware about sizing so this fails *loud and
early* instead of *silently at first boot*:

1. **Per-role soft recommendation, not just the global hard floor.**
   When the operator selects **Control plane** (or a control-plane HA
   member), check disk + RAM against the *recommended* numbers
   (~40 GiB disk / ~6–8 GiB RAM), and if under, show a whiptail
[...]

Issue #134, opened 2026-05-10
OS appliance: operationalize APPLIANCE.md + fill remaining gaps


mzac:
The design surface for the OS appliance lives in
[`docs/deployment/APPLIANCE.md`](https://github.com/spatiumddi/spatiumddi/blob/main/docs/deployment/APPLIANCE.md)
— base-OS choice (Alpine + Debian dual-track), image formats
(ISO / qcow2 / vmdk / ova / vhd / Docker), `mkosi` build pipeline,
first-boot wizard, update CLI, license analysis, security hardening.
That doc is the spec; this issue is the implementation epic — turn
it into bytes + close the gaps the doc doesn't yet answer.

Phase ordering and acceptance criteria are open — the questions in
each bucket below need decisions before code lands.

---

## Architectures

- [x] **amd64** — Phase 1 ships amd64 only.
- [ ] **arm64** (Phase 5) — required for Raspberry Pi 4 / 5 target
      (the agent containers already build multi-arch per
      non-negotiable #11). Builder image stays amd64; cross-arch
      mkosi runs use QEMU emulation.
- [ ] **RPi-specific**: build a separate "Raspberry Pi DHCP/DNS
      appliance" image with `dtb` + `firmware-rpi` packages,
      hardware-passthrough for physical NICs, GPU/HDMI excluded?
- [ ] **i386** — no. EOL.
- [ ] Per-architecture image matrix or one universal arm64 image
      with conditional firmware?

## Image / runtime model

- [ ] **One image for all services** (single VM = control plane +
      DNS + DHCP) vs **one image per role** (control-plane image +
      dns-appliance image + dhcp-appliance image)? Lean: ship both
      shapes — single image is easiest for first-time evaluators,
      role-split images are what real distributed deployments need
      (control plane in one DC, DNS/DHCP at edges).
- [ ] **Container runtime on the appliance** — Docker, Podman, or
      vanilla containerd? Implications: systemd-cgroup integration,
      log forwarding shape, restart policy semantics. APPLIANCE.md
      assumes "Docker images run natively" but doesn't pick one.
- [ ] **Containers vs native packages** — install Postgres / Kea /
      BIND9 as host-OS packages and run them via systemd, OR run
      the existing ghcr.io agent images on the appliance? Same
      image set across deployment paths is a strong argument for
      containers (already built, tested, signed).
- [ ] **Image tag vs immutable id** — appliance carries
[...]

Issue #549, opened 2026-07-03
Appliance: headless / unattended install — preseed the installer wizard via cloud-init user-data


mzac:
Broken out of #134 (OS appliance operationalization). The parent bullet:

> **First-boot wizard** is in APPLIANCE.md §4 — but it assumes the
> operator has console access. Add a **headless mode**: preseed via
> cloud-init user-data (network, admin password, role selection,
> control-plane URL for distributed roles).

## Problem

The install flow is now the interactive **whiptail** installer
`appliance/mkosi.extra/usr/local/bin/spatium-install` (post-#183 k3s
rewrite; Phase 8a disk-partitioning). It requires a VGA/serial console
to click through every screen. There is **no way to drive a fully
unattended install-to-disk** — a problem for fleet rollouts, IPMI/PXE
provisioning, cloud images, and CI boot-tests.

The existing `appliance/cloud-init/` NoCloud datasource (marked done in
#134 Phase 1) only covered the **superseded** pre-#183 flow: configuring
an already-running all-in-one over docker-compose. It does **not** feed
the current disk-partitioning installer. APPLIANCE.md §4 documents this
explicitly (`Phase 1 (pre-#183, replaced): headless via cloud-init
NoCloud`).

## Goal

On boot of the installer media, if a preseed **answer file** is present,
run `spatium-install` **non-interactively** with those answers instead
of launching whiptail. Partial preseed falls through to interactive
prompts only for the missing fields (so an operator can preseed
everything-but-the-disk, etc.).

## Fields to preseed (mirror the interactive wizard)

Every value `spatium-install` collects today:

- **Target disk** (the partitioning target — the destructive one; guard
  hard, see below)
- **Hostname** (RFC 1123 validated)
- **Admin user + password** (secret — see handling)
- **Network** — DHCP or static (IP / prefix / gateway / DNS)
- **Timezone** (default UTC)
- **Role** — `control-plane` (First node) vs `appliance` (Additional
  node / Application)
- **k3s pod + service CIDRs** (#302; defaults `10.42.0.0/16` /
  `10.43.0.0/16`)
- **Control-plane URL** — required when `role=appliance`
- **Bootstrap** — 8-digit pairing code (or PSK) when `role=appliance`

## Open design questions

1. **Datasource / transport.** Options, not mutually exclusive:
   - **NoCloud CIDATA ISO** — reuse `appliance/cloud-init/` shape; a
     second attached disk labelled `CIDATA` carries `user-data`
[...]

PR #579, opened 2026-07-06
feat(appliance): unattended installer — answer-file-driven headless install from the ISO (#134)

amoona6:
## Summary

Closes the headless gap in the installer ISO (#134): `spatium-install` now runs **completely non-interactively** when it finds a flat `key: value` answer file (`spatium-install.conf`), producing the same **6-partition A/B / STATE / HA-capable layout** as an interactive install — no console puppeting required.

Until now the two provisioning paths were mutually exclusive: the cloud-init **disk-image** path is headless but keeps mkosi's simple layout (no A/B slots, no STATE partition, no `/etc` overlay — so no slot upgrades, no HA promotion), while the **installer ISO** produces the real layout but could only be driven by hand through the whiptail wizard. This PR unifies them: headless *and* HA-capable.

## How it works

- **Discovery** (priority order): a `spatium.answers=<auto|/dev/X|/path|URL>` kernel-cmdline token (URL fetch covers PXE/netboot), else a block device labelled `SPATIUM`, else a NoCloud `CIDATA` seed carrying the file. The staged copy lives on tmpfs so the seed is released before partitioning. A new **"Install SpatiumDDI (unattended, from seed disk / URL)"** GRUB entry makes the intent explicit; the default Install entry auto-detects seeds too.
- **Parsing**: same dependency-free `sed` shape as `spatium-etc-render`'s `get_cfg`, plus inline-comment stripping and quote handling (quoted values keep `#` verbatim so passwords survive). No YAML library anywhere in the boot path.
- **Validation** mirrors every interactive prompt rule: role, hostname shape, admin user/password, static-network addresses (python `ipaddress`), pairing-code separator-strip + 8-digit check, k3s CIDR overlap incl. the static-LAN cross-check, and the 32 GiB A/B disk floor. The control-plane soft-sizing warning (#312) logs instead of blocking (no operator present to acknowledge it).
- **Safety**: the ERASE step is gated on a verbatim `confirm_wipe: true` — an accidentally-attached seed can never wipe a box. **Any** validation failure prints a full diagnostic (all errors, not just the first) and falls back to the interactive wizard; nothing is written. `target_disk: auto` picks the first disk over the floor.
- **`do_install` gains exactly two unattended seams** (the gauge sink becomes a console/log line filter; the Done msgbox becomes a summary + reboot) — all partitioning / rsync / chroot logic is shared verbatim with the interactive path, so the two paths cannot drift.
- **Offline pre-flight**: `spatium-install --check-answers <file>` lints an answer file on any machine (no root, touches nothing; machine-specific checks downgrade to warnings).

## Test plan

- [x] `appliance/tests/test_unattended_install.py` — 58 host-portable pytest cases driving `--check-answers`: the `confirm_wipe` consent gate (6 near-miss variants), role/hostname/credential/disk-floor/static-network/pairing/CIDR rules, parser behaviors (inline comments, quoting, `#` in quoted passwords), multi-error reporting, and two guards that keep the shipped example exactly one `confirm_wipe` flip away from valid. All pass; the pre-existing `appliance/tests` suites still pass (the 4 `test_slot_upgrade_runner` failures reproduce identically on `main` — macOS environment, unrelated).
- [x] `bash -n` + `shellcheck -S warning` on `spatium-install` + `wrap-iso.sh`: no new findings (the two pre-existing warnings are untouched).
- [x] Manual smoke of `--check-answers`: the shipped example refuses with exactly one error (`confirm_wipe`); flipping it validates OK.
- [ ] **Manual (not run here — needs a hypervisor + ISO build):** `make appliance-dev-iso`, then boot with a `genisoimage -volid SPATIUM` seed → zero-keystroke install → `lsblk` shows the 6-partition layout, web UI at the appliance IP; boot with **no** seed → interactive wizard unchanged; seed without `confirm_wipe` → non-destructive refusal; `role: appliance` seed → `pending_approval` in the Fleet tab; static-network seed → correct IP.

## Notes for review

- One deliberate deviation from the implementation plan sketch: `role` has **no default** in the unattended path (the plan's schema marks it required; silently defaulting the control-plane/appliance decision is the exact footgun the interactive wizard guards with a confirm dialog).
- Log-path env overrides + `/dev/null` fallbacks, the launch-log silencing, and the trace skip in check mode exist so `--check-answers` (and its pytest suite) run unprivileged on dev machines; live-ISO behavior is unchanged.
- Docs: new "Unattended install" section in `docs/deployment/APPLIANCE.md` §4, an "Installer ISO vs. disk image" callout in `appliance/cloud-init/README.md`, the stale `COMPOSE_PROFILES` comment in `user-data.example` fixed, field reference in `appliance/cloud-init/spatium-install.conf.example`, CLAUDE.md roadmap entry (🟡 pending review).

Design/plan context: research on the #134 headless gap + the implementation plan live in the local `thoughts/` tree (not committed).


mzac:
Thanks for this @amoona6 — genuinely thorough work, and the `--check-answers` offline linter + the 58-case pytest suite are excellent.

Closing as **superseded**, though: this overlaps almost entirely with #549 ("headless / unattended install via cloud-init preseed"), which merged in `eccf46f` and shipped in **2026.07.04-1**. This branch was cut before #549 landed, so it's a second implementation of the same headless-install feature — both drive `spatium-install` to the same 6-partition A/B layout, and #549 adds partial-preseed (fill some fields, prompt for the rest) on top. Carrying both would mean two rival mechanisms (`load_preseed`/`spatium-preseed.yaml` vs `discover_answer_file`/`spatium-install.conf`) in one installer.

Two things here are genuinely worth keeping that #549 lacks, though — the **offline linter** and the **test suite** (#549 shipped this disk-wiping path with no tests at all). I've opened #581 to port both onto #549, credited to you, and to fold in the parts of a security pass that apply to the live #549 code.

Sorry the timing collided — really appreciate the effort, and the linter idea in particular is going straight into #549.

Issue #625, opened 2026-07-12
chore(dhcp): smaller DHCP UX/behavior improvements (pool ordering, reserved badge, per-scope DNS-sync flag, default IP)


waza-ari:
## Summary

Tracking issue for **smaller DHCP UX / behavior improvements** landed alongside the agentless FortiGate DHCP work (#624). Grouped here so they can be reviewed/merged together rather than as scattered one-liners. More items will be appended to the checklist over time.

## Improvements

- [ ] **Display order of multiple dynamic pools.** When a scope has more than one dynamic pool, the pools render in an inconsistent order. Sort them deterministically (by start address) so the DHCP Pools view and the IPAM pool-boundary rows are stable and predictable.
- [ ] **"Reserved" badge in the DHCP Pools view.** Surface a badge for reserved pools (alongside dynamic / excluded) so operators can tell pool types apart at a glance in the pool list.
- [ ] **Per-scope flag to enable/disable DNS sync.** A scope-level toggle controlling whether the scope's addresses feed the IPAM↔DNS sync (so ephemeral / device-managed scopes don't generate or drift DNS records). Threaded through the scope create/edit form + API + `ScopeResponse`.
- [ ] **Default IP address when allocating a specific IP.** When the operator opens the "allocate specific IP" flow, pre-fill a sensible default (e.g. the next free address in the subnet) instead of an empty field, so the common case is one click.

## Notes

- Frontend-heavy (display order, reserved badge, default-IP prefill); the DNS-sync flag spans API + model + form.
- Keep each as its own commit within the feature branch so they can be split out if a reviewer wants them separate from the core FortiGate driver.

_This is a living checklist — additional small DHCP improvements will be appended as they're made._

Issue #616, opened 2026-07-11
dhcp: Windows DHCP write-through never fires on scope/subnet soft-delete — deleted scopes keep serving forever


mzac:
## Summary

On a DHCP server group backed by **Windows DHCP**, deleting a scope (or a subnet carrying scopes) from the UI removes it from SpatiumDDI's view and from the rendered Kea bundle — but **never removes it from the Windows DHCP server**, which keeps handing out the scope and all of its reservations indefinitely. There is currently *no* code path that reaches `push_scope_delete` for a normal operator delete.

Found while investigating a contributor question about static-reservation lifecycle. This one is independent of that design discussion and is a shipped data-plane bug.

## Root cause

`push_scope_delete` is only ever called on the `permanent=True` branch of both delete operations, and the soft path `return`s before reaching it:

- `_apply_delete_scope` — `backend/app/services/ai/operations_risky.py:838-851` commits and returns inside `if not args.permanent:`; `push_scope_delete(db, scope)` sits at `:854`, i.e. permanent-only.
- `_apply_delete_subnet` — same shape. Soft path returns at `operations_risky.py:283`; the `for scope in scope_rows: await push_scope_delete(db, scope)` loop is at `:302`, permanent-only.

And nothing else picks it up:

- The UI never passes `permanent=true` — `frontend/src/lib/api.ts` `deleteScope: (id) => api.delete(\`/dhcp/scopes/${id}\`)`.
- `permanent_delete_from_trash` (`backend/app/api/v1/admin/trash.py:259`) explicitly does **not** run the write-through hooks — its own docstring says so ("Unlike the per-resource `?permanent=true` flag (which both deletes *and* runs the legacy write-through hooks), this endpoint just removes the DB row").
- `trash_purge._sweep` (`backend/app/tasks/trash_purge.py:66-71`) issues a Core `delete(model)` — zero Python runs, so no write-through.

`grep -rn push_scope_delete backend/app/` returns exactly two call sites, both in the permanent branches.

## Impact

Agent-based (Kea) groups are fine: the global `deleted_at IS NULL` filter (`backend/app/db.py:189`) drops the soft-deleted scope from `config_bundle.py`'s `select(DHCPScope)`, so `subnet4` empties on the next poll and `collect_wake` converges agents in ~2s. Verified empirically against the dev stack.

Windows DHCP (an **agentless, authoritative-writer** driver) has no such pull loop — it only converges on an explicit push. So:

[...]

PR #630, opened 2026-07-12
feat(dhcp): agentless FortiGate cloud DHCP driver + scope-deletion cleanup (#623, #624, #625)

### LINKED ISSUE #623: fix(dhcp): scope deletion still strands dynamic leases + IPAM mirror rows (visible on the frontend) after #621
## Summary

After #621, deleting a DHCP scope **still** leaves dangling artifacts — and they are visible to the operator on the IPAM subnet page, not just in the raw DB:

1. **Static reservation IPAM mirrors persist.** The reservation rows themselves are correctly hidden by #617's soft-delete cascade, but every reservation also owns a *separate* `ip_address` mirror row (`status="static_dhcp"`, back-linked via `static_assignment_id`). That mirror is not released on the soft path, so the subnet IP table keeps showing those IPs as `static_dhcp` "Reservation" rows after the scope is gone.
2. **Dynamically-polled leases (FortiGate / Windows) persist and "cannot even be deleted."** They survive scope deletion entirely, keep their `dhcp` / `auto_from_lease` IPAM mirror, and there is no UI to remove them.
3. **Both inflate subnet utilization** — `_update_utilization` (`backend/app/api/v1/ipam/router.py:310-315`) counts every non-`available` row as allocated.

**Live evidence** (dev DB + operator screenshot on `10.10.20.0/24`):
- Active lease `10.10.20.150` (server `FG-70-1`, driver `fortigate`) still points at soft-deleted scope `f902bff4`; its `dhcp` / `auto_from_lease` mirror survives.
- `static_dhcp` mirrors `10.10.20.2` / `10.10.20.5` still point at soft-deleted statics.
- Subnet reports `5 / 254 allocated` — three of those five are the dangling mirrors above.

## Root cause

**Dynamic leases have no lifecycle tie to the scope.**
- `DHCPLease` (`backend/app/models/dhcp.py:753-795`) has no `SoftDeleteMixin` / no `deleted_at`.
- `DHCPLease.scope_id` FK is `ON DELETE SET NULL` (nullable).
- `_collect_descendants` (`backend/app/services/soft_delete.py:178-189`) has **no lease branch**; `DHCPScope` has **no `leases` relationship**. Leases are owned by the *server* (`server_id ON DELETE CASCADE`).
- So scope deletion — soft or permanent — never enumerates or removes lease rows. Soft-delete leaves a dangling `scope_id`; permanent nulls it; the lease survives either way.

**The static IPAM mirror is not released on the soft path, and even the permanent path's release is insufficient for the frontend.**
- #618's `detach_ipam_for_scope_statics` runs **only** on the permanent path (`operations_risky.py:920`), so a UI soft-delete (the default) releases nothing.
- Even the permanent path only *detaches to `available`* (`static_ipam.py:105-138`). But `list_addresses` (`router.py:6466`) returns all rows with no status filter, and the frontend paints one row per returned address (`IPAMPage.tsx:5029-5058`; "free" is the *absence* of a row; only `network`/`broadcast` are suppressed). An `available`-status row therefore still renders as a visible row. To fold the IP back into a free gap, the mirror `ip_address` **row must be deleted** — exactly what the lease-mirror cleanup already does (`dhcp_lease_cleanup.py:157`).

**Re-population + stranding.**
- `dhcp_pull_leases` iterates by server every ~10s and re-learns whatever the device reports, regardless of scope state — so even a manual API lease delete is transient while the source device still serves it.
- The zero-wire floor guard (`pull_leases.py:352-375`, #482) skips absence-delete on an empty wire response. So deleting a server's *only* scope makes its wire go empty and the leases are stranded until the time-based expiry sweep reclaims them (`expires_at`, hours/days later, if ever).

**"Cannot even be deleted" is literal on the frontend.** The backend endpoint `DELETE /dhcp/servers/{server_id}/leases/{lease_id}` exists (`servers.py:1389-1452`, SuperAdmin), but `frontend/src/lib/api.ts` `dhcpApi` has no `deleteLease` client method and the LeasesTab context menu offers only Copy actions.

## Why the previous commits didn't fix it

#621 (`b07415a`, closing #616–#619) is fully present in the tree (migration `b3e7d21c9f04` applied; DB at head `c9a1f4e07b52`). It fixed:
- #616 — agentless write-through now fires on both delete paths.
- #617 — pools + statics now ride the soft-delete cascade batch (this part works — the reservation *rows* are correctly hidden).
- #618 — IPAM mirror release, but **only on the permanent path**, and via *detach-to-`available`* (row kept) rather than deleting the row.
- #619 — `StaticUpdate` validation.

It **never addressed dynamic leases** at all, **never released the static mirror on the soft path**, and #618's detach-to-`available` still leaves a visible row. Hence the artifacts persist.

## Proposed fix (explored)

- **One reusable `purge_lease` helper** — DRY the duplicated teardown in `pull_leases.py:391-447` and `servers.py:1394-1452` (resolve subnet → revoke DDNS → delete `auto_from_lease` mirror → stamp `record_lease_history(…, "removed")` → delete lease), wired into **every** scope-reaching delete path: `_apply_delete_scope` (soft + permanent), the subnet/block/space cascade (driven off the soft-delete `batch.rows`), the permanent-subnet loop, and trash permanent-delete.
- **Delete DHCP-derived IPAM mirror rows** (static + lease) on scope delete so IPs fold back into free gaps and utilization drops (new `remove_ipam_for_static` / `remove_ipam_for_scope_statics` deleting the row + tearing down DNS). Re-create static mirrors on Trash restore (`upsert_ipam_for_static`); leases re-populate from the next poll (`push_scope_restore` re-creates the device scope).
- **Refine the zero-wire floor guard** so an empty wire when the server's group has **0 live scopes** runs absence-delete instead of skipping (closes the "deleted the only scope → leases stranded" gap). Reject an upsert-side guard — it would break intentional scopeless-subnet lease tracking.
- **Frontend** — add `dhcpApi.deleteLease` + a Delete item in the Leases tab; clarify the scope-delete modal to state the dynamic clients that will be released.
- **Data-repair migration** — one-shot cleanup (no schema change) deleting the already-stranded lease/mirror rows + `static_dhcp` mirrors pointing at soft-deleted statics.

Correctness constraints: purge **before** `apply_soft_delete` / `db.delete(scope)` (subnet CIDR must resolve; `SET NULL` must not fire first); single transaction, no inner commits; best-effort DDNS revoke; stamp history before deleting the lease row.

### LINKED ISSUE #624: feat(dhcp): agentless cloud DHCP driver family + FortiGate (FortiOS REST) implementation
## Summary

Add an **agentless** DHCP driver family that manages DHCP directly on devices already running their own DHCP server — over the device's REST API, with **no Kea agent deployed** — with **FortiGate / FortiOS** as the first concrete driver. This mirrors the shape of the agentless firewall integrations (#605 / #606) but for the DHCP subsystem, and extends the existing Windows-DHCP agentless (read-only) path into a fully bidirectional one.

## Motivation

Many networks already run DHCP on the perimeter device (FortiGate) or in the cloud. Today SpatiumDDI's DHCP management assumes a managed Kea agent (or the read-only Windows path). An agentless driver lets operators keep DHCP on the device while SpatiumDDI stays the source of truth: push scopes / pools / reservations to the device and pull its active leases back into IPAM — no per-site agent to deploy.

## Design

- **`AgentlessDHCPDriverBase`** (`backend/app/drivers/dhcp/_cloud_base.py`) — generic base for REST-managed DHCP devices (mirrors the cloud DNS base):
  - Whole-scope writes: `apply_scope_full(server, scope)` / `remove_scope_full(server, subnet_cidr)`.
  - Lease pull: `get_leases(server)`; `probe` / `health_check`; credential decrypt (`_load_credentials`).
  - `render_config` / `apply_config` raise `NotImplementedError`, `reload` / `restart` no-op (the device owns the daemon). Push-only — `get_scopes` is intentionally not implemented, so the pull path runs lease-reconcile only.
- **`FortiGateDHCPDriver`** (`backend/app/drivers/dhcp/fortigate.py`) — maps a neutral `ScopeDef` → FortiOS `system.dhcp.server`:
  - First-class fields always emitted for replace-all semantics: netmask (from CIDR), default-gateway, dns-server1..4, domain, ntp-server1..3, bootfile, lease-time.
  - Dynamic pools → `ip-range`; excluded/reserved pools → `exclude-range` clipped to the dynamic range; statics → `reserved-address` (no `action` field).
  - Interface CIDR matching (`_match_interface`, raises on no-match / ambiguous); leases via `/monitor/system/dhcp` (ipv4, `leased` / `reserved`).
  - Credentials: `api_token` / `vdom` / `verify_tls` (Fernet-encrypted at rest; token never echoed).
- **Registry** (`backend/app/drivers/dhcp/registry.py`) — `AGENTLESS_DRIVERS = {windows_dhcp, fortigate}`, new `CLOUD_DHCP_DRIVERS = {fortigate}` + `is_cloud()` helper. FortiGate is agentless but **not** read-only (`READ_ONLY_DRIVERS` stays `{windows_dhcp}`) — it supports `/sync` reconcile.
- **Write-through seam** (`backend/app/services/dhcp/cloud_writethrough.py`) — `push_cloud_scope_upsert` / `push_cloud_scope_delete`, integrated into the existing `windows_writethrough` push helpers so cloud/REST members fan out first, alongside Windows members, on every scope/pool/static change + delete + restore.
- **API** (`backend/app/api/v1/dhcp/servers.py`) — cloud credentials on server create/update (merge semantics: `None`=leave / `{}`=clear / dict=merge), `POST /test-fortigate-credentials`, `GET /{server_id}/fortigate-interfaces` (interface preflight annotated with matched scopes), and an `is_cloud(...)` branch in `sync_server` that reconciles every active scope in the group.
- **DNS** — per-scope `dns_track_dynamic_leases` toggle (migration `c9a1f4e07b52`) gating whether dynamically-polled leases feed the IPAM↔DNS drift check (`services/dns/sync_check.py`), so ephemeral leases don't read as "out of sync".
- **Tasks** — `dhcp_pull_leases` + `dhcp_health` pick FortiGate up automatically via the `AGENTLESS_DRIVERS` registry entry.

## Scope / follow-ups

**In scope:** push scopes/pools/statics to the device, pull active leases into IPAM, sync reconcile, credential test, interface preflight, offline driver unit tests (`test_fortigate_dhcp_driver.py`, `test_dhcp_fortigate_agentless_member.py`).

**Out / later:** per-lease *release on the device* (no seam today — the driver reads leases but never writes/releases them; a lease is only affected on-device by re-pushing or deleting the whole scope object); FortiManager JSON-RPC centralization; additional cloud DHCP providers on the same `AgentlessDHCPDriverBase`.

## Dependency

> ⚠️ Building this surfaced a pre-existing **scope-deletion lifecycle bug** — dynamic leases + IPAM mirror rows are stranded when a DHCP scope is deleted (filed separately). Since agentless drivers are exactly the kind that produce polled dynamic leases, the delete-workflow fix should land alongside or before this feature so those leases are cleaned up correctly on scope deletion.

waza-ari:
## Summary

This branch began as the **first agentless external DHCP driver (FortiGate)**, and wiring reservation lifecycle through it surfaced a set of scope-deletion cleanup bugs that survived #621. It bundles three related pieces:

- **Closes #624** — agentless FortiGate cloud DHCP driver (+ the reusable agentless cloud-driver base + write-through seam).
- **Closes #623** — scope deletion no longer strands dynamic leases, IPAM mirror rows, or DNS records.
- **Addresses #625** — the per-scope DNS-sync flag and the "Reservation" badge (the other tracker items — dynamic-pool ordering, default next-free IP — are **not** in this PR).

## #624 — Agentless FortiGate cloud DHCP driver
- `drivers/dhcp/{fortigate,_cloud_base}.py`: outbound FortiOS REST + Bearer API token, VDOM-scoped, no co-located agent — modeled on the agentless cloud DNS drivers. Whole-`system.dhcp.server`-object rebuild + PUT per scope; SpatiumDDI subnet CIDR → FortiGate interface matching; `reserved-address` for statics; `exclude-range` clipped to the dynamic range; FortiOS quirks handled (no `action:assign`, exclude-range within ip-range, no empty `domain`).
- New `CLOUD_DHCP_DRIVERS` / `is_cloud()`; `services/dhcp/cloud_writethrough.py` (`push_cloud_scope_upsert` / `push_cloud_scope_delete`) fans into the #621 `windows_writethrough` seam, so FortiGate rides every path incl. #616 soft-delete and restore.
- Servers API: cloud-credential handling + `/test-fortigate-credentials` + `/{id}/fortigate-interfaces` + a synchronous `/sync` full reconcile; frontend server picker + credential form.
- **Live E2E** against a lab FortiGate passed (push → soft-delete removes the device object → restore re-adds).

## #623 — Scope-deletion cleanup
#621 fixed the pool/reservation cascade, but a scope delete still left DHCP-derived rows dangling in the IPAM subnet view (and inflating utilization):
- The reservation's `status="static_dhcp"` IPAM **mirror** was never released on the soft path — and even the permanent path only freed it to `available`, which still renders as a visible row (the subnet table paints one row per `ip_address`; "free" is the *absence* of a row). It's now **deleted** so the IP folds back into a free gap.
- Dynamic **leases** (`DHCPLease` + their `auto_from_lease` mirror) were never touched — only a nullable `ON DELETE SET NULL` scope backlink — so they stayed fully live and couldn't be removed.

Fix: a shared `services/dhcp/lease_cleanup.py::purge_lease` / `delete_leases_for_scope` (DRY'ing three inline copies), `static_ipam.remove_ipam_for_*` + `remirror_scope_statics` on restore, wired into every scope-reaching delete path (scope / subnet / block / space cascade, group, permanent, trash-permanent) before the soft-delete stamp; a zero-wire floor-guard refinement; a DNS-sync reconcile that now clears **no-primary** stale records from the DB; a standalone per-lease Delete in the Leases tab; and a one-shot data-repair migration for installs already past #621.

## Verification
- Backend suite green (**2888 passed / 0 failed / 2 skipped**); `ruff` + `black` + `mypy` (733 files) clean; frontend typecheck / eslint / prettier clean.
- Migration verified on a live DB: 3 stranded rows → 0, IPs fold into free gaps, utilization drops; orphaned DNS records → 0.

## Note for maintainer — related DNS-side finding (not fixed here)
The agentless **cloud DNS** connectors (Cloudflare / Route 53 / Azure DNS / Google DNS, and the Windows-DNS path) likely have the **same soft-delete orphan bug class** fixed here for DHCP: soft-deleting a zone/record wouldn't retract it from the provider. Worth investigating on the DNS side. Caveat to weigh: cloud DNS providers have no rename, so `apply_zone_change` is delete-then-recreate — pushing a *delete* on soft-delete is safe / negligible-churn for **records**, but for **zones** it tears down + recreates provider-side (churns zone IDs + NS records). Suggested direction: fix records + restore unconditionally; make zone-delete-on-soft-delete an explicit decision.

## Note on commit structure
The three concerns share files (`api.ts`, `servers.py`) and chained migrations, so a clean per-issue commit split wasn't possible at whole-file granularity — this is two commits (the #623 fix; the FortiGate feature + #625 items).


backend/alembic/versions/d7b3f2a9c15e_repair_stranded_dhcp_lease_static_mirrors.py
```diff
@@ -0,0 +1,118 @@
...
+    #     — and a zone with no primary DNS server can't push a wire delete, so
+    #     they'd otherwise linger forever as un-cleanable "ip-deleted" stale rows
+    #     in the subnet's DNS view. GSLB pool records (``pool_member_id``) aren't
+    #     IPAM-backed, so they're excluded.
+    op.execute("""
+        DELETE FROM dns_record r
+        WHERE r.auto_generated = true
+          AND r.pool_member_id IS NULL
+          AND (
+            r.ip_address_id IS NULL
```
mzac: 🔴 **Critical / unrecoverable data loss.** This `r.ip_address_id IS NULL` arm makes the DELETE match *every* `auto_generated` record with no IP link — not just the DHCP mirror-orphaned ones. A null `ip_address_id` is the **normal** state for integration/ACME records:

- ACME DNS-01 challenge TXT — `services/acme_client/dns01.py:124-131`
- Kubernetes ingress — `services/kubernetes/reconcile.py:746-755`
- Tailscale mesh — `services/tailscale/reconcile.py:574-583`
- NetBird mesh — `services/netbird/reconcile.py:511`

All construct `DNSRecord(auto_generated=True, …)` with no `ip_address_id` and no `pool_member_id`, so every one is inside this predicate. On upgrade this wipes the entire Tailscale/NetBird mesh zones, all K8s ingress records, and any in-flight ACME challenge TXT — and `downgrade()` is a no-op, so it can't be undone.

Scope the delete to records this migration actually orphaned: capture the ids of records pointing at the `ip_address` rows removed in steps 1–2 before deleting them, then delete only those. The blanket `ip_address_id IS NULL` arm should go — it's not evidence of orphaning.

backend/app/drivers/dhcp/fortigate.py
```diff
@@ -0,0 +1,568 @@
...
+        Single seam tests patch to inject a fake transport — keep every
+        request flowing through the returned client. The API token rides
+        in the Authorization header (never the URL) and the VDOM is a
+        default query param on every call.
+        """
+        token = self._token(creds)
+        host = getattr(server, "host", "") or ""
+        port = int(getattr(server, "port", 443) or 443)
+        vdom = str(creds.get("vdom") or "root")
+        verify = bool(creds.get("verify_tls", False))
```
mzac: 🟠 **High — insecure TLS default.** This defaults certificate verification **off**, so the FortiOS admin Bearer token is sent over an unverified channel unless the operator opts in. Every sibling FortiOS integration defaults verify *on* (`services/fortinet/client.py`). Recommend `creds.get("verify_tls", True)` here, flipping the matching `setdefault("verify_tls", False)` calls in `servers.py` (create/update/probe) and `useState(false)` in `CreateServerModal.tsx:49`, and adding an optional `ca_bundle_pem` field like `FortinetClient` for private-CA FortiGates.

backend/app/drivers/dhcp/fortigate.py
```diff
@@ -0,0 +1,568 @@
...
+            else:
+                # Adopt + replace-all: overwrite whatever was on this
+                # interface's DHCP server with SpatiumDDI's desired state.
+                logger.info(
+                    "fortigate.adopt_existing_dhcp_server",
+                    server=str(getattr(server, "id", "")),
+                    interface=interface,
+                    mkey=existing_id,
+                )
+                resp = await client.put(f"/cmdb/system.dhcp/server/{existing_id}", json=body)
```
mzac: 🟠 **High — silently clobbers unmanaged FortiGate config.** `_find_server_id` matches an existing `system.dhcp.server` object purely by interface name, with no ownership/created-by marker, and this PUT then replaces its ip-range / exclude-range / reserved-address / options subtables with `_build_server_body`. If the operator already had a hand-managed DHCP server on that interface (custom pool, reservations, option 43), the first Sync overwrites it with no warning. Refuse to adopt a non-SpatiumDDI object unless explicitly opted in, or surface the existing object's range/reservation/option counts in the `/fortigate-interfaces` preflight so the destructive replace is visible first.

frontend/src/pages/dhcp/CreateServerModal.tsx
```diff
@@ -41,6 +42,13 @@ export function CreateServerModal({
   const [winUseTLS, setWinUseTLS] = useState(false);
   const [winVerifyTLS, setWinVerifyTLS] = useState(false);
   const [winClearCreds, setWinClearCreds] = useState(false);
+
+  // FortiGate-only fields. Same "blank = keep stored" contract on edit.
+  const [fgToken, setFgToken] = useState("");
+  const [fgVdom, setFgVdom] = useState(server?.vdom ?? "root");
+  const [fgVerifyTLS, setFgVerifyTLS] = useState(false);
```
mzac: 🟠 **High — edit silently disables TLS verify.** `fgVerifyTLS` starts `false` and is never hydrated from the server (the API doesn't echo `verify_tls`), yet the fortigate save branch unconditionally sends `verify_tls: fgVerifyTLS` in `cloud_credentials`, which the backend merges over the stored blob. So an operator who created the server with verify **on** and later reopens the modal for any unrelated edit (rename, change group, change VDOM) silently flips it back to **off**. Fix: echo `verify_tls` on `ServerResponse` (non-secret, like `vdom`) and `useState(server?.verify_tls ?? true)`, or only include `verify_tls` in the patch when the operator actually toggled it.

mzac (dismissed):
Thanks for this — the FortiGate driver is genuinely well-built (clean ABC implementation, the API token is Fernet-encrypted and never round-trips back out, and the push-failure → 502 → rollback contract is solid). I did a full review + security audit. **No malicious or unexpected code** — the only two URLs in the diff are the templated FortiOS base and a `wpad.dat` string used purely as test-fixture data; the WinRM-hardening test edit only stubs the new cloud push to a no-op and weakens no assertion; the `operations_risky.py` change doesn't widen the AI copilot's powers.

Requesting changes on one **critical** data-loss issue plus a few high-severity items before merge.

---

### 🔴 Critical — the `#623` repair migration deletes unrelated DNS records

Step 3b of `d7b3f2a9c15e` is an **unscoped** `DELETE` (inline comment below). The `r.ip_address_id IS NULL` arm has no correlation to the stranded DHCP rows the migration is repairing — it matches *every* `auto_generated` record with a null IP link, which is the **normal** state for integration- and ACME-synthesized DNS records. Confirmed producers: ACME DNS-01 challenge TXT (`services/acme_client/dns01.py:124-131`), Kubernetes ingress (`services/kubernetes/reconcile.py:746-755`), Tailscale mesh (`services/tailscale/reconcile.py:574-583`), NetBird mesh (`services/netbird/reconcile.py:511`) — all build `DNSRecord(auto_generated=True, …)` with no `ip_address_id` and no `pool_member_id`. So on upgrade this wipes the entire Tailscale/NetBird mesh zones, all Kubernetes ingress records, and any in-flight ACME challenge TXT, and `downgrade()` is a no-op → unrecoverable. Scope the delete to records that steps 1–3 actually orphaned (capture their ids first, or drop the blanket `IS NULL` arm). This migration also has no test — worth one that seeds a stranded DHCP record *and* an ACME/integration record and asserts only the former is removed.

### 🟠 High
- **TLS verification defaults OFF** (`fortigate.py:145` + `servers.py` setdefaults + `CreateServerModal.tsx:49`). A full read-write FortiOS admin Bearer token is sent over an unverified TLS channel by default — every other outbound-credential integration here (`services/fortinet/client.py`) defaults verify ON. Flip the default to `True` everywhere and consider a `ca_bundle_pem` field like `FortinetClient`.
- **Editing a server silently re-disables TLS verify** (`CreateServerModal.tsx:49`, inline). `fgVerifyTLS` is `useState(false)`, never hydrated (the API doesn't echo `verify_tls`), yet the edit unconditionally re-sends `verify_tls: false` — so any later edit (rename, move group, change VDOM) flips a previously-secured server back to unverified. Echo `verify_tls` on `ServerResponse` (it's non-secret, like `vdom`) and seed the checkbox from it.
- **First sync adopts + clobbers a pre-existing FortiGate DHCP server** (`fortigate.py:456`, inline). `_find_server_id` matches purely on interface name with no ownership marker, then PUTs a full `_build_server_body`, silently replacing an operator's hand-managed ip-range / reserved-address / options on that interface. Refuse adoption of a non-SpatiumDDI object unless explicitly opted in, or surface the existing object in the `/fortigate-interfaces` preflight.

### 🟡 Medium
- **Soft-delete → restore is lossy** (`static_ipam.py`). `remove_ipam_for_scope_statics` hard-deletes the reservation's `ip_address` mirror; restore recreates a minimal row, losing any operator-authored `description` / `tags` / `custom_fields` / `owner_user_id`. Breaks the soft-delete losslessness contract — preserve the row (or snapshot + re-apply the operator columns on restore).
- **API-created FortiGate server defaults to port 67** (`servers.py:125`). A server POSTed via the REST API without `port` gets 67, and the driver then dials `https://host:67` → every sync/lease/health call fails. Default cloud drivers to 443 when the caller omits `port` (the UI already sends 443).

### 🔵 Low
- No SSRF advisory guard (`assert_safe_target`) on the new outbound endpoints (`servers.py:747`), unlike the sibling integration test-connection endpoints.
- Numeric options (`mtu`, `time-offset`) are emitted as FortiOS type `string`, so clients get the ASCII text instead of the binary integer (`fortigate.py:421`).
- Force-Sync 502s are swallowed in the UI — no operator feedback (`DHCPPage.tsx` `syncMut` has no `onError`).
- New driver has no entry in `docs/drivers/DHCP_DRIVERS.md`.
- `test_dhcp_fortigate_agentless_member.py` monkeypatches `push_cloud_scope_upsert` / `build_scope_def` — the functions under test — so its member-detection assertions are vacuous. Drive at least one path through the real function, stubbing only the leaf `_client`.
- `detach_ipam_for_scope_statics` is now dead (all three call sites moved to `remove_ipam_for_scope_statics`) but still defined + exported.

---

**Housekeeping (I can push these to your branch if you'd like — `maintainerCanModify` is on):** the red check is just the migration-shape linter flagging the `#625` column-add's `downgrade()` drop (the standard safe same-release case — needs a one-line `scripts/lint_migrations.py --baseline` regen). And the branch is 1 commit behind `main`; it merges cleanly.

Separately, your maintainer note about the cloud DNS side was spot-on — I verified it and filed **#632** (it's real, High severity, and the 30-day purge makes it permanent; bulk record delete is already safe though).

backend/alembic/versions/d7b3f2a9c15e_repair_stranded_dhcp_lease_static_mirrors.py
waza-ari: Thanks, fixed in [5daa5ea](https://github.com/spatiumddi/spatiumddi/pull/630/commits/5daa5ea8113241760c80a9a4b082c53186be9920).

backend/app/drivers/dhcp/fortigate.py
waza-ari: Thanks, fixed in [5daa5ea](https://github.com/spatiumddi/spatiumddi/pull/630/commits/5daa5ea8113241760c80a9a4b082c53186be9920).

backend/app/drivers/dhcp/fortigate.py
waza-ari: Thanks, fixed in [5daa5ea](https://github.com/spatiumddi/spatiumddi/pull/630/commits/5daa5ea8113241760c80a9a4b082c53186be9920).

frontend/src/pages/dhcp/CreateServerModal.tsx
waza-ari: Thanks, fixed in [5daa5ea](https://github.com/spatiumddi/spatiumddi/pull/630/commits/5daa5ea8113241760c80a9a4b082c53186be9920).

mzac:
Merged as #635 — thank you, @waza-ari, this was a genuinely well-built
contribution. 🙏

**Why this PR is being closed rather than merged directly:** it isn't
anything about the code. This PR is opened from a fork, and GitHub only
creates the `dynamic` workflow runs that produce **Code Quality** and
**CodeQL** results for pull requests originating from branches *inside*
the repository — never for fork PRs. Our `main` ruleset has a
`code_quality` rule, so this PR was parked indefinitely at *"Code quality
results are pending for 2 analyzed languages"*, with no run in existence
to produce those results and nothing to re-trigger. It could not be
merged, and an admin bypass would have landed it with no static analysis
having run against it at all.

So I re-landed your **identical commits** (head `6abb36c`, authorship
preserved) on an in-repo branch as #635. The analyses then ran properly.
They surfaced exactly one finding — an `except (KeyError, TypeError,
ValueError): pass` with no explanatory comment in the FortiGate `mkey`
parse — which I fixed with a comment in `00afb83`, documenting why
swallowing is safe there (a ref with a missing/non-numeric `mkey` means
the scope was never pushed to that FortiGate, so there's nothing of ours
to claim, and leaving it out of `owned_mkeys` correctly reports any DHCP
server on that interface as *unmanaged* — we refuse to adopt rather than
clobber operator config). No other changes were made to your work.

The commits in `main` carry your authorship. Closes #623, #624 and #625.

Thanks again — and apologies for the friction; the fork/Code-Quality gap
is on our side to fix, and I'm looking into a ruleset change so the next
external contribution doesn't hit this.

PR #738, opened 2026-07-30
feat(dns): add Technitium DNS Server as a third DNS driver

Cmonnich:
## Summary

Adds Technitium as a fully agent-managed authoritative DNS driver alongside BIND9 and PowerDNS, wired end-to-end through the OS appliance:

- Control-plane driver (`backend/app/drivers/dns/technitium.py`) + agent driver (`agent/dns/spatium_dns_agent/drivers/technitium.py`) reconciling zones/records against Technitium's REST API, with agent-provisioned bearer-token auth (mirrors PowerDNS's local-API-key trust boundary).
- New `ghcr.io/spatiumddi/dns-technitium` container image (Ubuntu/glibc base — Technitium ships no Alpine package or release binaries).
- Docker Compose, CI (build matrix, release jobs, Trivy, kind e2e), and full appliance integration: supervisor role tables, firewall policy layer + seed migration, umbrella + appliance Helm charts, Fleet UI.
- **v1 scope**: primary zones + standard record CRUD. DNSSEC, native DoT/DoH/DoQ, catalog zones, secondary zones, and `dns_import` are tracked as fast-follow (see `docs/drivers/DNS_DRIVERS.md` §4B.5).

Also fixes two pre-existing bugs (not Technitium-specific) surfaced by live appliance testing on real hardware:

- `SupervisorCapabilities` is a strict Pydantic model that silently drops any capability flag it doesn't explicitly declare, despite its docstring's "ignores unknown keys" framing. `can_run_dns_technitium` was added on the supervisor side but never declared here, so the Fleet UI reported "supervisor doesn't advertise" even though the real supervisor was sending it. Added the field plus a regression test locking the full flag set.
- nginx's `client_max_body_size` override for large uploads only covered `/api/v1/backup/`, so the appliance upgrade-image upload endpoint (`/api/v1/appliance/upgrade-images`, backend-capped at 4 GiB) fell through to nginx's ~1MB default and 413'd every real slot-image upload. Added a matching override in all three nginx config copies (docker-compose, appliance ISO, Helm chart).

## Test plan

- [x] Unit tests: control-plane driver (14 tests), agent driver (21 tests), full agent suite (90 tests), full supervisor suite (252 tests) — all passing
- [x] Live docker-compose test against a real control plane: zone/record create/update/delete propagate to a real Technitium daemon and resolve via `dig`; agent cache survives a control-plane outage; recovers cleanly from a full 401/404 rebootstrap cycle
- [x] Real kind cluster e2e for both the umbrella chart and the appliance-specific chart (the exact DaemonSet template the ISO deploys) — real registration, token provisioning, and DNS resolution
- [x] 336 passing backend tests (appliance + firewall suites) after all supervisor/chart/firewall changes; clean frontend typecheck/lint
- [x] Trivy-clean container image (fixed a real vulnerable-runtime issue in the upstream base image along the way)
- [x] Built the actual appliance ISO + A/B slot image with `dns-technitium` baked in for air-gapped installs
- [x] Ran a fully automated, preseeded disk install inside QEMU — real GRUB A/B slot menu, k3s core came up Ready
- [x] Verified the two bug fixes live: role assignment for `dns-technitium` and a real 2.1 GB upgrade-image upload both confirmed working end-to-end on a real installed test appliance

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -59,6 +59,7 @@ updates:
       - "/agent/dhcp/images/kea"
       - "/agent/dns/images/bind9"
       - "/agent/dns/images/powerdns"
+      - "/agent/dns/images/technitium"
       - "/agent/dns/images/dnsdist"
       - "/agent/supervisor/images/supervisor"
       - "/agent/looking-glass/images/gobgp"
--- a/.github/workflows/agent-e2e.yml
+++ b/.github/workflows/agent-e2e.yml
@@ -1,15 +1,20 @@
 name: Agent E2E (kind)
 
-# Spins up a kind cluster, installs the umbrella Helm chart with both
-# DNS-agent flavors (BIND9 + PowerDNS) enabled side by side, and
-# verifies:
+# Spins up a kind cluster, installs the umbrella Helm chart with all three
+# DNS-agent flavors (BIND9 + PowerDNS + Technitium) enabled side by side,
+# and verifies:
 #   1. Every deployment + statefulset reaches Ready.
 #   2. The API's /health/live returns 200 via a port-forward.
 #   3. The BIND9 agent answers ``dig version.bind CH TXT`` — cheap
 #      "daemon is alive" probe that doesn't need pre-seeded zone data.
 #   4. The PowerDNS agent answers ``dig id.server CH TXT`` — pdns's
 #      CHAOS-class equivalent.
-#   5. Both pods register and stay stable (low restart count).
+#   5. The Technitium agent answers a query for the RFC 6761 reserved
+#      "invalid." TLD — Technitium REFUSES CHAOS-class queries entirely
+#      (confirmed empirically), so it has no id.server/version.bind
+#      equivalent; querying "invalid." is answered locally and instantly
+#      (NXDOMAIN, no recursion) regardless of zone state instead.
+#   6. All three pods register and stay stable (low restart count).
 #
 # Tag selection — control-plane vs agent images publish on different
 # cadences, so they need different tags:
@@ -97,7 +102,7 @@ jobs:
         run: |
           set -euxo pipefail
           docker buildx create --use --name e2e-builder >/dev/null 2>&1 || docker buildx use e2e-builder
-          for flavor in bind9 powerdns; do
+          for flavor in bind9 powerdns technitium; do
             docker buildx build \
               --platform linux/amd64 \
               --file "agent/dns/images/${flavor}/Dockerfile" \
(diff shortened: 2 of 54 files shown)


mzac:
Thanks for this — really thorough work, and the two pre-existing bug fixes (the strict-`SupervisorCapabilities` drop and the nginx `client_max_body_size` gap) were good catches that would have been easy to miss.

Closing in favour of #747, which is the same commit re-pushed as an in-repo branch. Nothing wrong with the contribution — it is a mechanical constraint: CodeQL `dynamic` runs are never created for fork PRs, so this repo's required `code_quality` check hangs pending forever and the PR can never merge. Your commit is preserved and you are credited via `Co-Authored-By`.

What changed on the way over:

**Merge with `main`** (17 commits had landed since you branched). Three doc conflicts, plus one git could not see — your firewall seed migration `6a668dd451d5` chained off `e5b2f09c71a4`, but `main`'s head has moved to `c9a4f2e81b56`. That merges silently and then breaks CI with two Alembic heads. Re-chained.

**Three correctness fixes in the record path.** The common root is that record CRUD bumps the config bundle's `etag` but not its `structural_etag` (`sync.py:172`), so `swap_and_reload`'s full-zone reconcile never fires on a record edit — which means nothing in that path self-heals:

1. `apply_record_op` sent `overwrite=false` unconditionally, and Technitium appends at that setting. Editing `www A 10.0.0.1` → `10.0.0.2` left the daemon serving both addresses round-robin, and a TTL-only edit came back `already exists` and was swallowed as an idempotent no-op. The docstring attributes this to PowerDNS parity, but PowerDNS's rrset `REPLACE` and BIND9's `upd.replace` both have neither problem. Now defaults to `overwrite=true`, honouring the `rrset_action="add"` override DNS pools use to avoid clobbering siblings. Your bulk reconcile correctly keeps `overwrite=false` — it computes an exact diff, so appending is right there.
2. The reconcile fingerprint excluded `ttl`, so it could not converge a TTL change either.
3. `invalid-token` (HTTP 200) had no recovery path. The state dir and `/etc/dns` are separate volumes and can desync, after which the agent wedged permanently.

Worth re-running your live docker-compose loop against fix 1 — it is reasoned from the API docs plus this repo's bind9/powerdns conventions and unit-tested, but not yet validated against a real daemon.

**Digest-pinned the base image.** `technitium/dns-server` is a third-party Docker Hub image that lands in the appliance ISO's baked set on every install, so a mutable tag means an upstream retag could change what ships with no diff in-repo. Pinned the multi-arch index digest so buildx still selects per-platform.

**Added 10 tests** over `apply_record_op`, `_reconcile_zones` and `_call`'s auth recovery — those paths had no coverage, and they are where all three bugs lived.

The deferred v1 items from `DNS_DRIVERS.md` §4B.5 now have tracking issues: #740 DNSSEC, #741 DoT/DoH/DoQ, #742 query-log shipping, #743 secondary/catalog zones, #744 `dns_import` + blocklists, #745 SVCB multi-value truncation. #746 is the umbrella. Contributions on any of them very welcome.

Issue #787, opened 2026-08-01
appliance(fleet): slot-image download 4XX — node-local image bytes with the mirror off on multi-node, orchestrator drops sha256/tls_insecure, and unchecksummed external-URL applies


mzac:
Field report (Discord, 2026-08-01):

> were you getting the **4XX error when trying to pull down the new image**? if so I used this commands to update the images via ssh
> ```
> sudo spatium-upgrade-slot apply https://github.com/spatiumddi/spatiumddi/releases/download/2026.07.30-1/spatiumddi-appliance-slot-2026.07.30-1-amd64.raw.xz --checksum https://…-amd64.sha256
> ```

The manual command **worked**. So the bytes and the GitHub URL are fine — the failure is in the control-plane-mediated path, which is structurally different from what the operator typed.

**The exact status code is not recoverable from the repo** — it is only written to the host's `/var/log/spatiumddi/slot-upgrade.log` (surfaced as `appliance.last_upgrade_log_tail`). This issue therefore does two things: names the diagnostic that settles it, and fixes the two **structurally unfixed** causes found while tracing, either of which produces a 4XX on a healthy install.

## Diagnostic that settles it

On the affected box, or from the DB row:

```bash
grep -i 'HTTPError\|urlopen\|http' /var/log/spatiumddi/slot-upgrade.log | tail -20
# or: SELECT last_upgrade_log_tail FROM appliance WHERE hostname='…';
```

The `urllib` `HTTPError` line carries both the status and the full URL, which separates every candidate below in one read.

## Why the fleet path differs from the manual command

The Fleet UI defaults to the **uploaded/imported image** source, not an external URL — `frontend/src/pages/appliance/FleetTab.tsx:4459`:

```tsx
const [sourceKind, setSourceKind] = useState<"url" | "uploaded">("uploaded");
```

In that mode `schedule_appliance_upgrade` (`backend/app/api/v1/appliance/supervisor.py:4904-4914`) hands the host a **control-plane** URL, not a GitHub one:

```
https://<control-plane>/api/v1/appliance/upgrade-images/<uuid>/raw.xz?t=<hmac>#a=<nonce>
```

That endpoint (`backend/app/api/v1/appliance/upgrade_images.py:830-895`) is the only 4XX-producing surface in the flow: `401` missing `?t=`, `403` HMAC mismatch, `404` row missing **or bytes missing on disk**.

[...]

Issue #392, opened 2026-06-12
Prune large appliance assets (ISO / slot raw.xz) from old GitHub releases — version-aware retention in CI


mzac:
## Summary

Every appliance release attaches ~**4.6 GB** of large binaries via
`action-gh-release` (`.github/workflows/release.yml`, the `files:`
block):

* a versioned ISO `spatiumddi-appliance-<ver>.iso` **and** a duplicate
  generic-named `spatiumddi-appliance-amd64.iso` (~1.1 GB each)
* a versioned slot `spatiumddi-appliance-slot-<ver>-amd64.raw.xz` **and**
  a duplicate generic `spatiumddi-appliance-slot-amd64.raw.xz` (~1.1 GB
  each)
* a handful of tiny `.sha256` sidecars

With **37 releases** today and ~10+ already carrying the full appliance
asset set, that's **~50 GB and growing ~4.6 GB per release** (sometimes
multiple a week). Nothing prunes it, so it grows unbounded and the
Releases page gets noisy. There's no GitHub storage *wall* for public-repo
release assets, so this is hygiene + clutter + bandwidth, not an outage —
but it's a clean, cheap win and it compounds.

## Two tiers of cleanup

**Tier 1 — drop the duplicate generic-named copies on NON-latest
releases (zero functional risk).** The generic names
(`…-amd64.iso`, `…-slot-amd64.raw.xz`) exist only to back the stable
`releases/latest/download/<name>` URLs, which always resolve to the
**newest** release. On any older release the generic copies are pure
duplication — the versioned copies already cover version-pinning. So the
generic copies can be deleted from every release except the latest with
no impact, immediately halving the heavy bytes per old release.

**Tier 2 — windowed retention on the versioned heavy assets.** Beyond a
keep window (e.g. last N releases or X days), delete the versioned
`.iso` + `.raw.xz` too. **Keep the tag, release, notes, and `.sha256`
sidecars always** — provenance is tiny and the binary can be rebuilt
from the tag if ever needed.

## The constraint — retention MUST be version-aware

The appliance OS-upgrade flow pulls **per-release** slot images at stable
URLs (`releases/<tag>/download/…` for a pinned target, plus
`releases/latest/download/…`), and the Fleet/OS-Versions/Rolling-Upgrade
UI offers operators a release picker (`applianceReleasesApi.list`). So
pruning a slot `.raw.xz` that the UI still offers would make "Apply
2026.05.xx-y" 404. Retention therefore has to:

[...]

Issue #386, opened 2026-06-12
Single-appliance OS upgrade: self-signed TLS download failure + silent crash-loop + no progress visibility


mzac:
## Summary

Applying an OS slot-upgrade to an appliance from an **imported / uploaded upgrade image** (the #199 flow) fails 100% of the time, retries silently forever, and surfaces nothing useful in the Fleet UI. Found while upgrading a single appliance (ddi1) to `2026.06.12-1`.

## Root cause (3 compounding bugs)

**A — download fails on self-signed TLS.** `apply_upgrade` builds `desired_slot_image_url` from `request.base_url` (`backend/app/api/v1/appliance/supervisor.py:4298`) → `https://<appliance>/api/v1/appliance/upgrade-images/<id>/raw.xz?t=…`, i.e. the appliance's **own self-signed** web cert. The host-side runner fetches it with bare `urllib.request.urlopen()` (`appliance/mkosi.extra/usr/local/bin/spatium-upgrade-slot:165`), which verifies TLS by default →

```
urllib.error.URLError: <urlopen error [SSL: CERTIFICATE_VERIFY_FAILED]
certificate verify failed: self-signed certificate>
```

Imported/uploaded images therefore never apply; only a direct public-CA GitHub URL would.

**B — silent crash-loop.** A failed apply renames the trigger to `.failed.<ts>` but leaves `desired_appliance_version` set, so the supervisor re-fires every heartbeat (~30 s) with no backoff and no clear-on-failure. 15+ `slot-upgrade-pending.failed.*` sidecars pile up in minutes. The **same anti-pattern** has produced **2189** `ntp-config-pending.failed.*` sidecars on ddi1 since the prior day — every host-config runner (slot-upgrade / ntp / snmp / firewall / tz) shares it.

**C — the Fleet UI hides all of it.** The Fleet drilldown (`frontend/src/pages/appliance/FleetTab.tsx:3710-3807`) shows only the coarse `last_upgrade_state` chip — no log, error, or progress. Worse, the failed→ready auto-heal resets the state the instant the trigger is renamed, so the API reports `last_upgrade_state=ready` *mid-failure*. (The local self-panel `SlotUpgradeCard.tsx` does carry a 120-line log tail; it's just never shown through Fleet.)

## Fix

[...]

PR #790, opened 2026-08-01
fix(appliance): slot-upgrade field bugs — clear-upgrade, rolling-path integrity hints, status version, asset URLs (#786, #787, #788, #789)

### LINKED ISSUE #786: appliance(fleet): "Clear failed upgrade" cannot clear a failed upgrade — host re-stamps the state every heartbeat, and the one heal path is gated behind the file that strands it
Field report (Discord, 2026-08-01, against `2026.07.30-1`):

> Failing on the upgrade already… not sure what's going on when trying to upgrade the appliance through fleet management. **The clear failed upgrade button doesn't do much.** It sends the POST to the clear-upgrade endpoint, but appliance remains in this error mode even after reboot.

Confirmed by code trace. The operator is right that the POST succeeds — it just doesn't clear any of the state that renders the error, and **no control-plane action can**, because the host is authoritative for those fields and re-stamps them every heartbeat. Not fixed on main.

## Root cause

`clear_appliance_upgrade` (`backend/app/api/v1/appliance/supervisor.py:4970-5007`) resets exactly four columns — all **desired**-state:

```python
row.desired_appliance_version = None          # :4987
row.desired_slot_image_url = None             # :4988
row.desired_slot_image_sha256 = None          # :4989
row.desired_slot_image_tls_insecure = False   # :4990
```

But the red "Upgrade failed" card is driven by four **observed**-state columns it never touches — `last_upgrade_state`, `last_upgrade_state_at`, `last_upgrade_progress`, `last_upgrade_log_tail` (`backend/app/models/appliance.py:513-533`). The Fleet UI keys on them directly:

```tsx
// frontend/src/pages/appliance/FleetTab.tsx:4472-4475
const showUpgradeStatus =
  upgradeInFlight || row.last_upgrade_state === "in-flight" || row.last_upgrade_state === "failed";
// :4235
const failed = state === "failed" || progress?.step === "failed";
```

So the "→ target version" chip disappears (desired cleared) while the failure card and its now-useless *Clear failed upgrade* button stay. Exactly "doesn't do much".

**Nulling those columns in the handler would not fix it either.** They are overwritten verbatim from the supervisor heartbeat on every tick (`supervisor.py:1655-1665`), which reads them from host sidecars under `/var/lib/spatiumddi/release-state/` — `slot-upgrade-pending.state`, `slot-upgrade.progress`, `slot-upgrade-pending.failed.<ts>`, `/var/log/spatiumddi/slot-upgrade.log`. **The host is the source of truth and there is no control-plane-driven command to reset it.**

**Why a reboot doesn't help:** every marker lives on the persistent `/var` partition, which the slot `dd` never touches — the #421 test docstring states this explicitly (`agent/supervisor/tests/test_slot_upgrade_stale_inflight.py:83-85`). The first heartbeat after boot re-reads `failed` and re-publishes it. Reboot is a no-op here by design.

## The one heal path is gated behind the thing that breaks

`clear_fleet_upgrade_marker()` (`agent/supervisor/spatium_supervisor/appliance_state.py:781-805`, called from `heartbeat.py:810-814` when the response carries no desired upgrade):

```python
_FIRED_URL_MARKER.unlink(missing_ok=True)
if _TRIGGER_FILE.exists():
    return                      # :796-797  ← silent hard gate
... if parts[0] == "failed": _HOST_SLOT_STATE.write_text("ready …")   # :798-804
```

Three properties make it fragile:

1. **Always one heartbeat late.** The tick delivering the clear has already published `failed` from `collect()`, so the UI cannot clear before the *following* tick.
2. **`if _TRIGGER_FILE.exists(): return`** — any failure that strands an unconsumed `slot-upgrade-pending` on disk wedges the heal permanently: the `.path`/`.service` unit failing to exec, the runner dying before its `trap on_exit EXIT` is armed (`spatiumddi-slot-upgrade:230` — e.g. `exec >> "$LOG"` failing at `:64-65`), or the stale-in-flight reaper's best-effort rename missing (`appliance_state.py:404-412`). The **same guard also blocks `maybe_fire_fleet_upgrade`** (`:745`), so the appliance can neither heal nor re-apply — a genuine deadlock with no operator escape from the UI.
3. Every `except OSError` on that path is silent.

## Two surfaces disagree, which compounds it

The per-box `/appliance` panel **does** self-heal `failed → ready` once the trigger is gone (`backend/app/services/appliance/slot.py:207-247`, auto-heal at `:234-235`) and deletes a stale `.state` before writing a new trigger (`:319-332`). The Fleet path deliberately dropped both behaviours in #386 (`appliance_state.py:298-318`) and never grew an equivalent. So the same appliance can read **ready** on its own page and **Upgrade failed** in Fleet, simultaneously.

## Suggested fix

- [ ] `/clear-upgrade` must also clear observed state, and — because the host wins — must **command the host to reset it**, not just write the DB. Natural shape: a `clear_upgrade_state` desired-flag on the heartbeat response that makes the supervisor remove `slot-upgrade-pending*`, reset `.state`/`.progress` to `ready`, and prune `.failed.<ts>` sidecars; the DB clear then sticks because the next heartbeat reports `ready`.
- [ ] Remove or invert the `if _TRIGGER_FILE.exists(): return` gate for the *explicit clear* case — an operator-initiated clear should delete a stranded trigger, which is precisely the state needing recovery. (Keep the guard for the passive no-desired-state path if it's protecting an in-flight apply; distinguish the two.)
- [ ] Give the Fleet surface the same trigger-absent auto-heal the per-box surface has (`services/appliance/slot.py:234-235`), so the two stop disagreeing.
- [ ] Log the silent branches (`except OSError`, the trigger-exists early return) so `kubectl logs` shows *why* a clear didn't take.

## Acceptance

- [ ] Clicking **Clear failed upgrade** returns the appliance to a clean state within one heartbeat, and it stays clean (no re-stamp from the host).
- [ ] A stranded/unconsumed `slot-upgrade-pending` is recoverable entirely from the UI — no SSH.
- [ ] Fleet and the per-box `/appliance` panel never report different upgrade states for the same box.
- [ ] Tests: none exist today for `clear_fleet_upgrade_marker`'s heal path (only a comment reference at `test_slot_upgrade_stale_inflight.py:78`), and none assert `/clear-upgrade` clears the visible error. Add both.

## Related

Same subsystem as #386 (silent crash-loop / failed→ready masking), #419 + #420 (stuck "Upgrade pending"), #421 (permanent `in-flight`), #551 (rollback safety-nets). Companion to the download-failure issue filed alongside this one. The `version=unreadable` report from the same thread is separate and cosmetic.

---

Filed together from the same Discord thread: #786 (this), #787 (slot-image download 4XX), #788 (`status` shows the active slot unreadable), #789 (release artifacts silently not published).

### LINKED ISSUE #788: appliance: `spatium-upgrade-slot status` always reports the ACTIVE slot as version=unreadable (RO mount of the already-mounted root)
Field report (Discord, 2026-08-01) after a manual slot upgrade to `2026.07.30-1`:

> now that I have the latest images, I see a discrepancy
> ```
> Active slot:    root_b    /dev/nvme0n1p5  version=unreadable
> Inactive slot:  root_a    /dev/nvme0n1p4  version=2026.07.21-1
> ```
> but info in the UI reads the version.

**The upgrade succeeded and the UI is correct.** `spatium-upgrade-slot status` reports the *active* slot as `unreadable` on every appliance, on every release, always — the operator is seeing a permanent display bug at the exact moment it looks like an upgrade failure.

## Root cause

`cmd_status` probes **both** slots with the same mount-based helper (`appliance/mkosi.extra/usr/local/bin/spatium-upgrade-slot:149-158`):

```python
print(f"Active slot:    {active['partlabel']:8}  {active['device']:14}  version={slot_version_label(active)}")
print(f"Inactive slot:  {inactive['partlabel']:8}  {inactive['device']:14}  version={slot_version_label(inactive)}")
```

and `slot_version_label()` unconditionally mounts the device read-only (`:110-123`):

```python
try:
    run(["mount", "-o", "ro", slot["device"], str(mountpoint)])
except subprocess.CalledProcessError:
    return "unreadable"
```

The active slot's device is **already mounted read-write at `/`** — every GRUB entry emits `root=UUID=… rw` (`spatium-grub-render:272,280,291,301`). The kernel reuses the existing superblock and refuses the differing `SB_RDONLY` flag; `mount(8)` short-circuits with "already mounted". Either way the helper returns the literal string `unreadable`.

So whichever slot is currently booted always prints `unreadable`. Before this operator's upgrade, `root_a` would have shown it and `root_b` a real version — the same bug, mirrored. (Their inactive-slot read succeeded, so this is not a missing-`sudo` artifact.)

## The rest of the codebase gets it right

`sync_slot_versions()` (`:652-682`) reads the active slot from the live `/etc/spatiumddi/appliance-release` with **no mount** via `_read_active_appliance_version()` (`:634-649`), and mounts only the inactive slot. It writes `/var/lib/spatiumddi/release-state/slot-versions.json`, which is what the supervisor heartbeat (`agent/supervisor/spatium_supervisor/appliance_state.py:1287`), the Fleet chips (`frontend/src/pages/appliance/FleetTab.tsx:3954-3964`) and the console dashboard (`spatium-console:1452-1468`) all read.

This matches the documented design — `docs/deployment/APPLIANCE.md:1096-1099`: *"Active slot reads its own `/etc/spatiumddi/appliance-release` directly; inactive slot is probed via a quick read-only mount."* **`cmd_status` is the only place that violates it.** So here the UI is authoritative and the host CLI is wrong — the reverse of the usual assumption, which is what makes the report confusing.

There is stale-comment evidence this was the original intent (`:627-629`): *"Updated by `sync_slot_versions()` called from `cmd_apply` and `cmd_status`"* — `cmd_status` has never called it (`sync_slot_versions()` is invoked only at `:365` from `cmd_apply` and `:692` from `cmd_sync_versions`).

## Blast radius: cosmetic

Nothing consumes `status` output programmatically, and nothing in the upgrade machinery branches on per-slot versions:

- commit-on-healthy-boot uses grubenv `saved_entry` vs booted PARTLABEL (`spatiumddi-firstboot:1035-1045`) — version-free
- `rollback` / `set-next-boot` / `set-default` (`:723`, `:777`, `:810`) — partlabel + grubenv only
- the fleet upgrade health gate keys on `installed_appliance_version` (the live-file read), not slot versions (`backend/app/services/upgrades/per_node.py:432-470`)
- `slot_a_version` / `slot_b_version` appear only in the model, the API response and the Fleet UI
- GRUB menu titles treat `unreadable` as a sentinel and read the (correct) sidecar (`spatium-grub-render:86`)

The real cost is operator trust: `status` is the command people run to confirm an upgrade landed, and it reports the one slot they care about as broken.

**Operator workaround today:** `sudo spatium-upgrade-slot sync-versions` prints both slots correctly.

## Not a regression

`git diff 2026.07.21-1 2026.07.30-1 -- …/spatium-upgrade-slot` is empty — the script is byte-identical across both tags; last touched in `b72f84e3` (2026-07-03). Unfixed on main.

## Fix

- [ ] Have `cmd_status` call `sync_slot_versions()` and print from the returned map — which its own comment at `:628` already claims it does. Alternatively branch `slot_version_label()` on "is this the booted root" and use `_read_active_appliance_version()`.
- [ ] Correct the stale comment at `:627-629`.

## Adjacent latent bug (higher severity, worth its own look)

`/etc` is an overlayfs with `upperdir=/var/persist/etc` on the **persistent** `/var`, shared across slot swaps (`spatium-install:1383-1438`). Both `_read_active_appliance_version()` and the supervisor's `read_installed_version()` read *through* that overlay. Directory copy-up is non-opaque, so today the new slot's lower `appliance-release` stays visible — but if anything ever writes or removes `/etc/spatiumddi/appliance-release`, the old version would be pinned in the upper layer forever across slot swaps. That would poison `installed_appliance_version`, which **does** gate the fleet upgrade health gate (`per_node.py:469`) — a hung upgrade, not a cosmetic glitch.

Minor: the cleanup comment at `:146` says the leftover probe dir "is under `/tmp` and tmpfs reclaims on reboot", but the mountpoint is `/run/spatium-slot-probe-*`.

---

Filed together from the same Discord thread: #786 ("Clear failed upgrade" no-op), #787 (slot-image download 4XX), #788 (this), #789 (release artifacts silently not published).

### LINKED ISSUE #789: docs(appliance): per-tag slot-image URLs use the stable asset name, which #392 prunes from every non-latest release (404)
Found while investigating two field reports about failed Fleet upgrades (#786, #787). Not the cause of either, but a real publishing gap with its own operator impact: **the release workflow never verifies that its appliance artifacts actually published**, and they frequently don't.

## Finding 1 — the "stable" alias is missing from every release except the newest

`.github/workflows/release.yml:586-595` has published each slot image under two names since `481a57a0` (2026-05-12, #141):

```bash
VERSIONED_XZ="…/spatiumddi-appliance-slot-${VERSION}-amd64.raw.xz"
STABLE_XZ="…/spatiumddi-appliance-slot-amd64.raw.xz"
mv "$SRC_XZ" "$VERSIONED_XZ"
cp "$VERSIONED_XZ" "$STABLE_XZ"
```

In practice only `2026.07.30-1` carries the stable pair. Audited across every release tag — `2026.06.12-1` through `2026.07.21-1` all have versioned assets only. Verified by HTTP probe:

| URL | Status |
|---|---|
| `releases/download/2026.07.21-1/spatiumddi-appliance-slot-amd64.raw.xz` | **404** |
| `releases/download/2026.07.21-1/spatiumddi-appliance-slot-2026.07.21-1-amd64.raw.xz` | 206 |
| `releases/download/2026.07.30-1/spatiumddi-appliance-slot-amd64.raw.xz` | 206 |
| `releases/latest/download/spatiumddi-appliance-slot-amd64.raw.xz` | 206 |

`latest/download/…` resolves **only because the newest release happens to have the alias** — it 404'd for the whole period before `2026.07.30-1` was cut. The workflow code was unchanged throughout, so this is an upload-reliability problem (two identical ~1.5 GB blobs per release, plus two identical ~1.5 GB ISOs), not a code change. Nothing in the job fails when an asset silently doesn't land.

## Finding 2 — some releases shipped with no appliance artifacts at all

Every release from `2026.05.12-1` back through `2026.05.12-2`, and `2026.05.13-1` / `2026.05.14-1` / `2026.05.17-*` / `2026.05.18-1` / `2026.05.22-*` / `2026.05.26-1` / `2026.06.04-1` / `2026.06.11-1`, carries **only the tiny `.sha256`** — no ISO, no `.raw.xz`. e.g.:

```
$ gh release view 2026.06.11-1 --json assets
spatiumddi-appliance-slot-2026.06.11-1-amd64.sha256   0MB
```

The pattern — the 118-byte sidecar lands, the 1.1–1.5 GB payloads don't — is consistent with large-asset upload failures that never failed the job. Those releases are un-installable and un-upgradable-to, and nothing surfaced that at cut time. (Some of that range predates parts of the Phase 8 work, so a few may be expected; the ones from `2026.05.12-1` onward are not, since the build step existed by then.)

## Finding 3 — code comments and docs assert the alias always exists

Three places state the stable name is always available, which is false for every release but the newest:

- `backend/app/services/appliance/releases.py:45-50` — *"Both names point at the same bytes; either works for download."*
- `backend/app/api/v1/appliance/upgrade_images.py:110-113` — same claim, and `_GITHUB_IMAGE_FILENAME = "spatiumddi-appliance-slot-amd64.raw.xz"` is stamped as the stored `filename` for **every** GitHub-imported image regardless of which asset was actually fetched, so the Fleet picker displays a name that doesn't match the release.
- `docs/deployment/APPLIANCE.md:1037-1038` — the OS Image card is *"pre-filled with `…/releases/latest/download/spatiumddi-appliance-slot-amd64.raw.xz` so a first-time operator just clicks Apply"*.

And the air-gap recipe at `docs/deployment/APPLIANCE.md:1268-1278` tells operators to `wget` a **pinned-tag stable-name** URL:

```
wget https://github.com/spatiumddi/spatiumddi/releases/download/2026.06.01-1/spatiumddi-appliance-slot-amd64.raw.xz
```

That shape has never resolved for any tag. Same for the release-notes body the workflow itself generates (`release.yml:768-784`), which advertises the stable URL in every release.

Not affected: the release-catalog picker resolves assets from GitHub's own `browser_download_url` and prefers the versioned name (`releases.py:164-181`), so the in-product "pick from GitHub Releases" flow builds correct URLs. It does require a matched `.raw.xz` + `.sha256` pair, so the Finding-2 releases are silently omitted from the dropdown — correct behaviour, but it hides a broken release rather than reporting it.

## Suggested fix

- [ ] **Verify after publish.** Re-query the release's assets at the end of the job and fail if any expected artifact is absent or its size is 0 / implausibly small. A release that silently ships without an ISO is worse than a failed workflow run.
- [ ] **Stop uploading the payload twice.** Two identical 1.5 GB blobs per artifact doubles upload time and failure surface for no capability — `releases/latest/download/…` already resolves the versioned name if consumers use the catalog. Either drop the stable copy and fix the consumers/docs that reference it, or keep exactly one and make the other a documented redirect.
- [ ] **Fix the three assertions in Finding 3** (two code comments + the pre-filled-URL and air-gap doc examples) so they describe whichever convention survives.
- [ ] Consider backfilling the stable alias onto recent still-supported tags, or explicitly document versioned-only as the contract.

## Acceptance

- [ ] A release that fails to publish any appliance artifact fails the workflow.
- [ ] Every documented download URL resolves for the tag it names, verified in CI rather than by hand.
- [ ] The GitHub-imported upgrade-image row records the asset name actually downloaded.

---

Filed together from the same Discord thread: #786 ("Clear failed upgrade" no-op), #787 (slot-image download 4XX), #788 (`status` shows the active slot unreadable), #789 (this).

mzac:
Four appliance slot-upgrade issues from a Discord field report on `2026.07.30-1`, filed as #786–#789. Every fix is verified on real hardware (ddi1) — see the validation comments below.

## What was reported

> The clear failed upgrade button doesn't do much. It sends the POST to the clear-upgrade endpoint, but appliance remains in this error mode even after reboot.

> were you getting the 4XX error when trying to pull down the new image? […] now that I have the latest images, I see a discrepancy — `Active slot: root_b … version=unreadable`

## What shipped

**#786 — "Clear failed upgrade" could not clear a failed upgrade.** The endpoint reset the four `desired_slot_image_*` columns, but the red card renders from `last_upgrade_state` / `last_upgrade_progress`, which the heartbeat re-publishes verbatim from host sidecars every ~30 s. The host owns that state, so clearing the DB was undone on the next heartbeat — and rebooting never helped, because those sidecars live on the persistent `/var` a slot write never touches.

The clear is now a command to the host: a fire-once `clear_upgrade_requested` flag (migration `e3b7a1c25d94`) that makes the supervisor reset its sidecars and drop a stranded `slot-upgrade-pending` trigger. Unlike the passive heal it is *allowed* to remove that trigger, because the existing `if _TRIGGER_FILE.exists(): return` guard is exactly what deadlocked wedged boxes — it also blocks `maybe_fire_fleet_upgrade`, so an appliance could neither clear nor re-apply.

**A permission bug meant none of that would have worked, and probably explains the original report.** `release-state` is mode 1777 (sticky), the sidecars are written by the *root* host runner as 0644, and the supervisor container runs `su-exec spatium`. A non-owner gets `EACCES` overwriting and `EPERM` unlinking — verified empirically, not by inspection. So the supervisor-side reset was inert, and by the same mechanism so is the long-shipped `clear_fleet_upgrade_marker`. The runner now publishes both sidecars `0666`, the supervisor overwrites `.progress` rather than unlinking it (unlink can never work under the sticky bit), and firstboot backfills the mode so an upgraded appliance repairs itself.

**#787 — the rolling orchestrator re-created #386.** It stamped only version + URL, never `desired_slot_image_sha256` / `desired_slot_image_tls_insecure`, so an uploaded image was fetched from the appliance's own self-signed URL with verification on and no hash; a stale sha256 from an earlier per-box schedule would fail the new image as corrupt. Both scheduling surfaces now go through one resolver (`services/appliance/slot_image_target.py`) that returns and stamps all four together. The `#a=` re-fire nonce moved there too and is now **run-scoped**, so an orchestrator resume no longer re-drives a node with a URL it never fired.

**#788 — `spatium-upgrade-slot status` always showed the active slot as `unreadable`.** It probed both slots by mounting them read-only, but the active slot is already mounted rw at `/`, so the kernel refuses the differing-flags mount. Whichever slot was *booted* always displayed as unreadable, on every appliance and every release — so the reporter's upgrade had in fact succeeded. Now reads the active version from the live `appliance-release`. `status` stays read-only; an earlier revision of this branch routed it through `sync_slot_versions()`, which would have let a probe taken mid-`dd` persist `unreadable` over the sidecar that feeds the Fleet card, the console and the GRUB titles.

**#789 — docs advertised a URL shape that 404s.** The air-gap recipe pinned a tag *and* used the un-versioned asset name; that pair resolves only while the tag is newest, because #392's retention prunes the un-versioned assets from every non-latest release. Fixed there and in the Rolling Upgrade URL placeholder, plus two code comments that claimed the two names are interchangeable.

## Deliberately not fixed

**The multi-node slot-image gap (#787, stays open).** Uploaded bytes live on whichever api replica served the upload; the mirror that fixes it is off by default and the appliance chart never enables it, so on multi-node the host's download 404s from every other replica. An earlier commit on this branch added a schedule-time refusal — **reverted in `23e7d2c0`**, because it asked the wrong question (appliance-node count, when the invariant is api replica count), so it missed the stock multi-replica plain-k8s default while falsely refusing single-api-replica appliance fleets; it made a blocking kubeapi call from an async handler; and its only remedy names a Helm value an appliance operator has no supported way to set, which would have left air-gapped multi-node fleets with **no** upgrade path. The hazard is documented in `APPLIANCE.md` and `values.yaml` instead.

**External-URL checksums (#787, stays open).** Reasoning is on the issue: `.xz` verifies its own container during decompress, so corruption already fails the apply cleanly, and fetching `<url>.sha256` from the host that served the image buys little against tampering — while the two proper implementations cost an SSRF surface or a trigger-format change across mixed-version fleets.

## Testing

`make ci` green. 217 appliance host-script tests, 280 supervisor tests, 91 targeted backend tests. New suites: `appliance/tests/test_slot_status_active_version.py`, `agent/supervisor/tests/test_clear_upgrade_command.py`, `backend/tests/test_slot_image_target.py`, `backend/tests/test_appliance_clear_upgrade.py` — all wired into CI. For #788 I confirmed the tests catch the bug by stashing the fix and watching three of them fail.

The full backend suite was **not** run locally (`-n auto` OOMs an 8-core/7 GB box, per CLAUDE.md) — this PR is how it gets exercised.

## Both draft blockers are discharged

1. **Host changes field-validated on ddi1.** The permission bug was reproduced from first principles (supervisor runs uid=100; overwriting a root-owned `0644` sidecar in the sticky `release-state` gives EACCES, unlink gives EPERM), the fix verified, and the whole clear flow driven end to end against real images: stranded trigger removed, `.state` reset, DB settled to `ready`, failure did not return. #788 likewise reproduced (`version=unreadable`) and fixed. Details in the two validation comments.

2. **#786 now retires the command on acknowledgement, not a timer** (`121f670f`). The stopwatch could expire before the command was ever delivered and needed two follow-up fixes; the host's compliance was observable all along, since the clear's job is to turn `failed` into `ready`. Validated on ddi1 with a deliberately 5-minute-stale stamp — which the old timer would have discarded on arrival — and the command was still delivered and applied.

**Two bugs were found by field testing that no test had caught**: a failed trigger unlink still reported success (`b8dc805c`), and the ingest suppression swallowed the host's own acknowledgement. Both fixed with regression tests.

Closes #786
Closes #788
Closes #789
Refs #787

### DIFF
--- a/agent/supervisor/spatium_supervisor/appliance_state.py
+++ b/agent/supervisor/spatium_supervisor/appliance_state.py
@@ -845,16 +845,31 @@ def apply_clear_upgrade_command() -> bool:
             # box stays wedged — exactly the failure this exists to fix.
             log.warning("supervisor.clear_upgrade.unlink_failed", path=str(path), error=str(exc))
 
+    # Only rewrite a state that is actually stuck. Two reasons: a healthy
+    # box would otherwise report "cleared" on every heartbeat inside the
+    # flag's lifetime, and clobbering a live ``in-flight`` stamp would blind
+    # BOTH crash detectors that key on that literal — the runner's exit trap
+    # and the supervisor's stale-in-flight reaper — so a mid-apply Cancel
+    # could leave a truncated image with nothing recording the failure.
     try:
         if _HOST_SLOT_STATE.exists():
-            _HOST_SLOT_STATE.write_text(f"ready {datetime.now(UTC).isoformat()}\n", encoding="utf-8")
-            reset = True
+            current = _HOST_SLOT_STATE.read_text(encoding="utf-8").split(maxsplit=1)
+            if current and current[0] == "failed":
+                _HOST_SLOT_STATE.write_text(
+                    f"ready {datetime.now(UTC).isoformat()}\n", encoding="utf-8"
+                )
+                reset = True
     except OSError as exc:
         log.warning("supervisor.clear_upgrade.state_reset_failed", error=str(exc))
 
+    # OVERWRITE rather than unlink. release-state is mode 1777 (sticky) and
+    # this file is root-owned (the host runner writes it); a non-owner can
+    # never unlink it, but the runner now publishes it 0666 so an in-place
+    # write succeeds. An empty object reads as "no progress" to every
+    # consumer, which is what a cleared upgrade should show.
     try:
         if _SLOT_UPGRADE_PROGRESS.exists():
-            _SLOT_UPGRADE_PROGRESS.unlink(missing_ok=True)
+            _SLOT_UPGRADE_PROGRESS.write_text("{}\n", encoding="utf-8")
             reset = True
     except OSError as exc:
         log.warning("supervisor.clear_upgrade.progress_reset_failed", error=str(exc))
--- a/agent/supervisor/spatium_supervisor/appliance_state.py
+++ b/agent/supervisor/spatium_supervisor/appliance_state.py
@@ -808,6 +808,61 @@ def clear_fleet_upgrade_marker() -> None:
         pass
 
 
+def apply_clear_upgrade_command() -> bool:
+    """#786 — the operator pressed "Clear failed upgrade"; make the host forget.
+
+    ``clear_fleet_upgrade_marker`` above is the *passive* heal, and it
+    deliberately returns early when a trigger file is still on disk so it
+    can't disturb an apply that is genuinely running. That guard is also
+    what made a wedged appliance unrecoverable: any failure that strands an
+    unconsumed ``slot-upgrade-pending`` — the path/service unit failing to
+    exec, the runner dying before its EXIT trap is armed, a missed rename
+    by the stale-in-flight reaper — blocks both the heal AND
+    ``maybe_fire_fleet_upgrade``, so the box could neither clear nor
+    re-apply, and no operator action could break the deadlock.
+
+    This is the explicit counterpart: the operator has *asked* for the
+    stranded state to go, so the trigger is removed rather than deferred
+    to. Returns True when anything was actually reset (for the log line).
+
+    Deliberately does NOT stop a running apply: it only removes the
+    trigger + resets the sidecars. A live ``spatiumddi-slot-upgrade``
+    process re-stamps its own progress as it goes, so a clear raced
+    against a real apply loses to the apply rather than corrupting it.
+    """
+    if detect_deployment_kind() != "appliance":
+        return False
+
+    reset = False
+    for path in (_TRIGGER_FILE, _FIRED_URL_MARKER):
+        try:
+            if path.exists():
+                path.unlink(missing_ok=True)
+                reset = True
+        except OSError as exc:
+            # Surfaced rather than swallowed: if the stranded trigger
+            # survives, the operator's clear silently didn't take and the
+            # box stays wedged — exactly the failure this exists to fix.
+            log.warning("supervisor.clear_upgrade.unlink_failed", path=str(path), error=str(exc))
+
+    try:
+        if _HOST_SLOT_STATE.exists():
+            _HOST_SLOT_STATE.write_text(f"ready {datetime.now(UTC).isoformat()}\n", encoding="utf-8")
+            reset = True
+    except OSError as exc:
+        log.warning("supervisor.clear_upgrade.state_reset_failed", error=str(exc))
+
+    try:
+        if _SLOT_UPGRADE_PROGRESS.exists():
+            _SLOT_UPGRADE_PROGRESS.unlink(missing_ok=True)
+            reset = True
+    except OSError as exc:
+        log.warning("supervisor.clear_upgrade.progress_reset_failed", error=str(exc))
+
+    _prune_failed_sidecars(keep=0)
+    return reset
+
+
 def _prune_host_config_sidecars(trigger_file: Path, keep: int = 5) -> None:
     """(#387) Keep only the newest ``.failed.<ts>`` / ``.done.<ts>`` /
     ``.invalid.<ts>`` sidecars for ONE trigger family, so a (pre-fix)
@@ -819,7 +874,10 @@ def _prune_host_config_sidecars(trigger_file: Path, keep: int = 5) -> None:
         parent = trigger_file.parent
         for suffix in ("failed", "done", "invalid"):
             stale = sorted(parent.glob(f"{trigger_file.name}.{suffix}.*"))
-            for path in stale[:-keep]:
+            # ``stale[:-0]`` is the EMPTY slice, not the whole list, so a
+            # plain slice would make keep=0 ("remove them all", used by the
+            # #786 explicit clear) silently prune nothing.
+            for path in stale if keep <= 0 else stale[:-keep]:
                 path.unlink(missing_ok=True)
     except OSError:
         # Best-effort housekeeping — leftover sidecars are cosmetic; a
--- a/agent/supervisor/spatium_supervisor/appliance_state.py
+++ b/agent/supervisor/spatium_supervisor/appliance_state.py
@@ -875,6 +875,21 @@ def apply_clear_upgrade_command() -> bool:
         log.warning("supervisor.clear_upgrade.progress_reset_failed", error=str(exc))
 
     _prune_failed_sidecars(keep=0)
+
+    # A trigger we could not remove leaves the box exactly as wedged as it
+    # was — it still blocks both the passive heal and maybe_fire_fleet_
+    # upgrade. Field-observed: a root-owned trigger (this directory is
+    # sticky, so a non-owner can never unlink one) failed here while the
+    # state/progress resets succeeded, and the command still reported
+    # success. Say so instead, or the operator is told a clear worked when
+    # the appliance is still stuck.
+    if _TRIGGER_FILE.exists():
+        log.error(
+            "supervisor.clear_upgrade.trigger_survived",
+            path=str(_TRIGGER_FILE),
+            hint="not owned by the supervisor; remove it from the host to unwedge",
+        )
+        return False
     return reset
 
 
--- a/agent/supervisor/spatium_supervisor/heartbeat.py
+++ b/agent/supervisor/spatium_supervisor/heartbeat.py
@@ -794,6 +794,16 @@ def heartbeat_once(
     desired_next_boot_slot = body_out.get("desired_next_boot_slot")
     desired_default_slot = body_out.get("desired_default_slot")
     reboot_requested = bool(body_out.get("reboot_requested"))
+    clear_upgrade_requested = bool(body_out.get("clear_upgrade_requested"))
+
+    # #786 — an explicit "Clear failed upgrade" outranks everything below.
+    # Handled BEFORE the desired-upgrade branch so a clear issued while a
+    # failed desired-state is still on the row resets the host in the same
+    # heartbeat, and before the passive heal because only this path is
+    # allowed to remove a stranded trigger file.
+    if clear_upgrade_requested:
+        if appliance_state.apply_clear_upgrade_command():
+            log.info("supervisor.heartbeat.clear_upgrade_applied")
 
     if desired_version and desired_url:
         if appliance_state.maybe_fire_fleet_upgrade(
--- a/agent/supervisor/tests/test_clear_upgrade_command.py
+++ b/agent/supervisor/tests/test_clear_upgrade_command.py
@@ -1,0 +1,103 @@
+"""#786 — the operator-driven "make the host forget a failed upgrade" command.
+
+The visible failure state is re-published from these host sidecars on
+every heartbeat, so clearing the DB alone never worked: the next
+heartbeat restored it, and rebooting didn't help because release-state
+lives on the persistent /var a slot write never touches.
+
+The passive heal (``clear_fleet_upgrade_marker``) can't cover the wedged
+case either — it returns early whenever a trigger file is present, which
+is exactly the state a stranded apply leaves behind, and that same guard
+also blocks ``maybe_fire_fleet_upgrade``. So a box could neither clear
+nor re-apply.
+
+``apply_clear_upgrade_command`` is the explicit counterpart: the operator
+asked, so the stranded trigger goes.
+"""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+
+import pytest
+
+from spatium_supervisor import appliance_state
+
+
+@pytest.fixture
+def paths(monkeypatch: pytest.MonkeyPatch, tmp_path: Path) -> dict[str, Path]:
+    state = tmp_path / "slot-upgrade-pending.state"
+    trigger = tmp_path / "slot-upgrade-pending"
+    progress = tmp_path / "slot-upgrade.progress"
+    marker = tmp_path / "slot-upgrade-fired-url"
+    monkeypatch.setattr(appliance_state, "_HOST_SLOT_STATE", state)
+    monkeypatch.setattr(appliance_state, "_TRIGGER_FILE", trigger)
+    monkeypatch.setattr(appliance_state, "_SLOT_UPGRADE_PROGRESS", progress)
+    monkeypatch.setattr(appliance_state, "_FIRED_URL_MARKER", marker)
+    # The command is appliance-only; every other deployment kind no-ops.
+    monkeypatch.setattr(appliance_state, "detect_deployment_kind", lambda: "appliance")
+    return {"state": state, "trigger": trigger, "progress": progress, "marker": marker}
+
+
+def _wedged(paths: dict[str, Path]) -> None:
+    """The shape a stranded failure leaves on disk."""
+    paths["state"].write_text("failed 2026-08-01T00:00:00+00:00\n", encoding="utf-8")
+    paths["progress"].write_text(json.dumps({"step": "failed", "pct": 40}), encoding="utf-8")
+    paths["trigger"].write_text("https://cp.local/x.raw.xz\n", encoding="utf-8")
+    paths["marker"].write_text("https://cp.local/x.raw.xz\n", encoding="utf-8")
+
+
+def test_clear_resets_every_sidecar_and_removes_the_trigger(paths: dict[str, Path]) -> None:
+    _wedged(paths)
+
+    assert appliance_state.apply_clear_upgrade_command() is True
+
+    assert not paths["trigger"].exists(), "a stranded trigger must be removed, not deferred to"
+    assert not paths["marker"].exists()
+    assert not paths["progress"].exists()
+    assert paths["state"].read_text(encoding="utf-8").startswith("ready ")
+
+
+def test_clear_removes_the_trigger_the_passive_heal_refuses_to_touch(
+    paths: dict[str, Path],
+) -> None:
+    """The regression this exists for: with a trigger on disk the passive
+    heal returns early and the box stays wedged forever."""
+    _wedged(paths)
+
+    appliance_state.clear_fleet_upgrade_marker()
+    assert paths["state"].read_text(encoding="utf-8").startswith("failed"), (
+        "passive heal is expected to bail while a trigger exists"
+    )
+
+    appliance_state.apply_clear_upgrade_command()
+    assert paths["state"].read_text(encoding="utf-8").startswith("ready ")
+
+
+def test_clear_prunes_every_failed_sidecar(paths: dict[str, Path]) -> None:
+    """#386's runaway re-fire left piles of these; an explicit clear takes
+    them all, not all-but-five."""
+    for ts in range(9):
+        (paths["trigger"].parent / f"slot-upgrade-pending.failed.{ts}").write_text("x")
+
+    appliance_state.apply_clear_upgrade_command()
+
+    assert list(paths["trigger"].parent.glob("slot-upgrade-pending.failed.*")) == []
+
+
+def test_clear_is_idempotent_on_an_already_clean_host(paths: dict[str, Path]) -> None:
+    """Nothing to reset → False, so the heartbeat doesn't log a reset that
+    didn't happen. Must not raise on the missing files."""
+    assert appliance_state.apply_clear_upgrade_command() is False
+
+
+def test_clear_is_a_noop_off_appliance(
+    paths: dict[str, Path], monkeypatch: pytest.MonkeyPatch
+) -> None:
+    monkeypatch.setattr(appliance_state, "detect_deployment_kind", lambda: "docker")
+    _wedged(paths)
+
+    assert appliance_state.apply_clear_upgrade_command() is False
+    assert paths["trigger"].exists()
+    assert paths["state"].read_text(encoding="utf-8").startswith("failed")
--- a/agent/supervisor/tests/test_clear_upgrade_command.py
+++ b/agent/supervisor/tests/test_clear_upgrade_command.py
@@ -55,8 +55,22 @@ def test_clear_resets_every_sidecar_and_removes_the_trigger(paths: dict[str, Pat
 
     assert not paths["trigger"].exists(), "a stranded trigger must be removed, not deferred to"
     assert not paths["marker"].exists()
-    assert not paths["progress"].exists()
+    # Overwritten, not unlinked: release-state is sticky and this file is
+    # root-owned, so a non-owner can never unlink it.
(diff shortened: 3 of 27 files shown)


mzac:
## Field validation on a real appliance (ddi1, `dev-2680a4d2-ae37`)

The "host-side changes are unvalidated" caveat in the description is now **partly discharged**. Three claims were verified on hardware rather than by inspection.

### 1. The #786 permission bug is real, and reproduces exactly

`release-state` is `drwxrwxrwt` (1777, sticky). Files written by the root host runner are `root:root 0644`; the supervisor container writes as **uid=100 gid=101** (visible on disk as e.g. `apt-config-pending.fire-state` owned `_chrony:_ssh`, versus `apt-config-pending.state` owned `root:root`).

Creating a root-owned `0644` `.state` and acting as the supervisor's uid:

```
--- as supervisor uid, try to OVERWRITE (the #786 reset) ---
sh: 1: cannot create …/ZZtest.state: Permission denied
--- as supervisor uid, try to UNLINK ---
rm: cannot remove '…/ZZtest.state': Operation not permitted
--- file still says: ---
failed 2026-08-01T00:00:00+00:00
```

So the supervisor-side reset was inert, exactly as the commit claims — and so, by the same mechanism, is the long-shipped `clear_fleet_upgrade_marker`. That is very likely the original field report's root cause.

### 2. The fix resolves it, and the overwrite-not-unlink choice is required

With the runner's new `0666` publish applied to the same file, the supervisor uid writes it successfully (`ready 2026-08-01T21:20:00+00:00`) — while `rm` is **still** refused by the sticky bit. That is precisely why the supervisor overwrites `.progress` instead of unlinking it; an unlink-based reset could never have worked.

Full chain exercised: simulate a pre-fix appliance (`root:root 0644`) → run the firstboot backfill loop → modes become `0666` → supervisor uid performs the reset. All four steps confirmed.

### 3. #788 reproduces and is fixed

Shipped binary on the real box:

```
Active slot:    root_a    /dev/sda4       version=unreadable
Inactive slot:  root_b    /dev/sda5       version=dev-2680a4d2-ae37
```

…while `slot-versions.json` correctly held `slot_a: dev-2680a4d2-ae37` — confirming the CLI was wrong and the UI authoritative. After hot-deploying this branch's `spatium-upgrade-slot`:

```
Active slot:    root_a    /dev/sda4       version=dev-2680a4d2-ae37
Inactive slot:  root_b    /dev/sda5       version=dev-2680a4d2-ae37
```

The sidecar's mtime is unchanged across two consecutive `status` runs, confirming it is read-only again; `sync-versions` still writes it.

### Still not validated

The **container side** of #786 — the `clear_upgrade_requested` flag flowing API → heartbeat response → supervisor → host reset — needs the new api + supervisor images deployed to k3s and a genuinely wedged upgrade to clear. Everything below that boundary is now proven; the flag plumbing is covered only by unit tests.

### Appliance state

Left with this branch's `spatium-upgrade-slot` and `spatiumddi-slot-upgrade` hot-deployed (both strict improvements); `spatiumddi-firstboot` untouched. No test residue, node `Ready`, all pods `Running`.

mzac:
## End-to-end validation on ddi1 — container side now covered

Built this branch's api + supervisor images, imported them into k3s on the appliance, ran the migration, rolled both workloads, and drove a real wedged upgrade through the clear.

**Migration.** `e3b7a1c25d94` applied cleanly against the live appliance database:

```
Running upgrade a4f1c93d7e28 -> e3b7a1c25d94, Host-side clear-upgrade command (#786)
```

Worth noting for anyone repeating this: `kubectl set image` must also update the **`wait-for-migrate` init container**, not just `api`. Left on the old image it waits for an alembic head that image doesn't know about and the pod sits in `Init:0/1` forever, while the old api pod correctly goes `0/1` because its schema-head readiness gate now sees a DB ahead of it.

**Baseline — the wedged state reproduces.** With a stranded `slot-upgrade-pending` (path unit stopped so nothing consumes it) and `.state` = `failed`, the passive heal bails as designed and the failure is stuck:

```
host state: failed 2026-08-01T21:35:00+00:00
DB says:    failed | clear_requested=false
```

**The clear works end to end.** After setting the flag, the next heartbeat delivered it and the supervisor reset the host:

```
supervisor.heartbeat.clear_upgrade_applied
  .state:   ready 2026-08-01T21:33:17+00:00
  trigger:  REMOVED
  marker:   REMOVED
DB:         ready | clear_requested=false
```

The stranded trigger and its fire-once marker are gone, the state is reset, and the DB settled to `ready` with the flag auto-expired after delivery — i.e. the failure did **not** come back, which is the whole point of the issue.

**A new bug surfaced, and is fixed in `b8dc805c`.** An earlier run used a *root-owned* trigger. The state and progress resets succeeded, but:

```
supervisor.clear_upgrade.unlink_failed
  error="[Errno 1] Operation not permitted: .../slot-upgrade-pending"
supervisor.heartbeat.clear_upgrade_applied      ← reported success anyway
```

`release-state` is sticky, so a trigger the supervisor doesn't own can never be unlinked — and a surviving trigger leaves the box exactly as wedged, because it still blocks both the passive heal and `maybe_fire_fleet_upgrade`. The command now returns False and logs an actionable error naming the file. Normal operation is unaffected (the supervisor writes the trigger, so it owns it), but the failure is no longer silent. Regression test added.

**Appliance restored**: sidecars removed, `spatiumddi-slot-upgrade.path` re-started and `active`, migrate Job deleted, image tarballs cleaned. 9/9 pods healthy, both slots reporting their version correctly.

Every claim in this PR is now verified on hardware except the operator-facing HTTP path itself (`POST /clear-upgrade`), which is unit-tested; the clear was driven by writing the same columns that endpoint writes.

mzac:
## Ack rework validated on ddi1 (`121f670f`)

The 15 s stopwatch is gone. The command now rides every heartbeat response until the host reports a state other than `failed` — which is exactly what a successful reset produces, so compliance is observed rather than assumed.

Decisive test: fired a clear with `clear_upgrade_requested_at` set **5 minutes in the past**. Under the old timer that command was already expired on arrival and would have been silently dropped — the exact lost-command failure that needed two follow-up patches. Under the ack design:

```
=== supervisor ===
supervisor.heartbeat.clear_upgrade_applied
=== host ===
  .state:  ready 2026-08-01T21:50:54+00:00
  trigger: REMOVED
=== DB: ready clear=false
```

Delivered, applied, and retired only once the host confirmed.

What the rework removes:

- **The lost-command class entirely.** A supervisor that is offline, restarting, or slow still gets the command; there is no window to miss.
- **The clock-skew hazard.** No timestamp comparison between appliance and control plane is involved, so a drifting appliance clock cannot strand the flag.
- **The swallowed acknowledgement.** Suppression narrowed from "every `last_upgrade_*` field for the whole window" to "only a report still carrying the dismissed `failed`", so the host's post-reset `ready` is ingested normally — and so is progress from an upgrade retried straight after clearing, which is the normal recovery gesture and was previously blanked for a full interval.

`clear_upgrade_requested_at` survives only as a give-up backstop for a host that can never comply (a trigger it does not own is unlinkable out of the sticky directory); hitting it logs a warning naming the state the host is stuck on, rather than failing silently.

Four new backend tests cover ack-retirement, keeps-riding-until-compliance, the give-up backstop, and the retry-progress case. Appliance restored: sidecars removed, path unit `active`, 9/9 pods healthy. CI 24/24.

Issue #455, opened 2026-06-25
Paginate large list endpoints (DNS zone records + DHCP leases) — server-side, 100/page


mzac:
## Problem

List endpoints that can grow unbounded return their whole result set in one query +
one payload, with no server-side pagination. On a large zone / lease table this makes
the DB poll and the UI render slow.

### DNS records (the trigger)
[`GET /dns/groups/{gid}/zones/{zid}/records`](backend/app/api/v1/dns/router.py) →
`list_records` does `select(DNSRecord).where(zone_id==…).order_by(name, type)` with
**no limit** and `response_model=list[RecordResponse]`. A 20k-record zone (e.g. the
perf seed in #452/#454) returns all 20k rows every poll. The group-level "all records
across the group's zones" view has the same shape.

### DHCP leases (same class of problem)
[`GET /dhcp/servers/{id}/leases`](backend/app/api/v1/dhcp/servers.py) → `list_leases`
is **limit-capped** (`limit=500`, hard max 5000) and ordered `last_seen_at desc`, but
there's **no true pagination** — no `offset`/`page`, no `total`. So you can't browse
past the most-recent N leases, and the UI can't show "page 2 of N". On a busy server
the older leases are simply unreachable through the API.

## Proposed fix

Server-side pagination, **default 100/page**:

- Return a paginated envelope `{items, total, page, page_size}` (⚠️ breaking
  response-shape change from the bare list — update every consumer).
- Add a server-side `search` filter (records: name/value/type; leases: ip/mac/hostname)
  so users can find a row without paging through.
- Frontend: page controls + search box on the DNS records table and the DHCP leases
  table; show `total`.
- Keep a **"fetch all"** path only where it's actually needed (zone export renders the
  full zone via the export endpoint, not this list endpoint — verify it doesn't depend
  on `list_records`).
- Update MCP record/lease list tools + the perf harness if it reads these endpoints
  (the seeder writes records via bulk-create, doesn't list them; phase0 lists with a
  tiny page — fine).

## Notes
- The DNS records performance was the visible trigger after the #454 bulk-seed work
  (zones now reach 20k+ records quickly).
- Consider whether the envelope should be shared (a generic `Page[T]`) so future list
  endpoints adopt it consistently.

Issue #186, opened 2026-05-16
[feat] UI should expose backend error messages in a better way


waza-ari:
### Area

Frontend / UI

### What problem does this solve?

Currently, when something fails, the UI often only shows a generic error message, even though a better message is provided by the backend. For example, trying to generate an appliance pairing code might just show 409:

<img width="913" height="489" alt="Image" src="https://github.com/user-attachments/assets/2e11f50f-63d7-4c8f-b0ec-45a99539cce9" />

The API does return a better message though which is visible on browser console or when issuing the same request using the API. Those should be surfaced to the UI

```json
{
  "detail": "No DNS_AGENT_KEY configured on the control plane. Set it in /etc/spatiumddi/.env and restart the api before issuing codes."
}
```

### Proposed solution

Surface error details to UI 

### Would you be willing to help?

- [ ] I'd be open to submitting a PR for this
- [x] I can help test a prototype

mzac:
Did you check under Administration -> Insights & Audit -> Diagnostics to see if the error shows up there?  The containers are supposed to catch errors and file them there.

waza-ari:
I did, however it does not show up there. There are two different (unrelated) errors I'm still looking into (how I managed to cause them), but this particular one is not showing up.

Also, side note: while the built-in admin user can access Administration -> Insights & Audit -> Diagnostics, an OIDC backed user with superadmin privileges cannot (403). See #190

Issue #31, opened 2026-05-03
OPNsense (tier 1 — firewall-of-choice for labs)


mzac:
- ⬜ **OPNsense** *(tier 1 — firewall-of-choice for labs).*
  Per-`OPNsenseRouter` row, API key + secret auth over HTTPS.
  REST endpoints: `/api/dhcpv4/leases/searchLease`,
  `/api/dhcpv4/settings/getReservation`,
  `/api/diagnostics/interface/getInterfaceConfig`,
  `/api/interfaces/vlan_settings/get`. Mirror LAN / VLAN /
  OPT* interfaces → Subnets, DHCP leases → IPAddress
  (`status="dhcp"`, same shape as Kea mirror), static mappings →
  `status="reserved"`. ARP table endpoint as a secondary
  population source for devices with static-IP-outside-DHCP.

mzac:
**Triage cross-link (2026-06-07) — decide together:** #32

_OPNsense and pfSense are explicitly paired tier-1 firewall integrations sharing the same reconciler; build/spec together._

PR #835, opened 2026-08-06
fix(api): admit the string 422 detail the API has always returned

### LINKED ISSUE #834: Hand-raised HTTPException(422, detail="…") returns a body the OpenAPI document declares invalid (270 sites)
## Hand-raised `HTTPException(422, detail="…")` returns a body the API's own OpenAPI document declares invalid

FastAPI auto-declares a `422` response for every operation that validates a parameter, pointing at `HTTPValidationError`, whose `detail` **must be an array** of `ValidationError`. That is honest for the 422s FastAPI's own request validation produces.

But **270 call sites across `backend/app/api/v1`** — 57 router files, led by `ipam/router.py` (24), `dns/router.py` (22) and `cutover/router.py` (18) — raise `HTTPException(status_code=422, detail="…")` by hand for semantic validation a signature cannot express. FastAPI serialises those as `{"detail": "<string>"}`.

The document says array, the wire says string, and nothing reconciles them. One status code, two incompatible body shapes, only one documented.

Nobody authored the offending declaration — FastAPI generated it, and the API has always behaved differently.

### Impact

A generated client that types `detail` per the spec breaks on every one of these responses. So does any consumer doing `for err in body["detail"]` — it iterates the characters of a sentence, or throws.

### Reproduce

```bash
curl -sk -H "Authorization: Bearer $TOKEN" \
  'https://<appliance>/api/v1/ipam/addresses/search?limit=128&sort=null'
```

```
HTTP/1.1 422 Unprocessable Content
{"detail":"sort must be one of: address, description, hostname, last_seen, mac, status"}
```

Against the same appliance's `/api/openapi.json`:

```json
"HTTPValidationError": {
  "type": "object",
  "properties": {
    "detail": { "type": "array", "items": { "$ref": "#/components/schemas/ValidationError" } }
  }
}
```

The same endpoint returns the *documented* shape for a model-level failure, so both really do occur:

```json
{"detail":[{"type":"missing","loc":["body","name"],"msg":"Field required"}]}
```

Found by a schemathesis 4.24.3 `response_schema_conformance` pass over the live 1086-operation document:

```
schemathesis.openapi.checks.JsonSchemaError: Response violates schema

  "sort must be one of: address, description, hostname, last_seen, mac, status"
      is not of type "array"

  Validated against the response schema for status code 422.
  Schema at /components/schemas/HTTPValidationError/properties/detail:
      { "type": "array", "items": { "$ref": "#/components/schemas/ValidationError" } }
```

### Scope

On one build, **29 of 39 red conformance rows were this single defect**. Re-running the full affected set on a nested QA appliance:

```
27 failed, 67 passed, 992 deselected in 326.54s
```

Grouping every violation instance in that run:

```
54  Validated against the response schema for status code 422
54  Schema at /components/schemas/HTTPValidationError
```

Unanimous — one cause, no exceptions. Affected operations span `ipam`, `bgp`, `asns`, `dnsbl`, `looking-glass`, `webhooks`, `new-devices`, `dns-threat`, `address-sets`, `circuits`, `bacnet`, `appliance/tls`, and the `dhcp`/`dns`/`ipam` import routers. A sample of distinct violating bodies, all bare strings:

```
{"detail":"At least one field must be provided in changes"}
{"detail":"asn must be a 32-bit unsigned integer (0 .. 2^32 - 1)"}
{"detail":"invalid classification"}
{"detail":"kind must be one of ['1to1', 'hide', 'pat']"}
{"detail":"limit must be between 1 and 500"}
{"detail":"space_id is not a valid UUID"}
{"detail":"strategy must be 'sequential', 'random', or 'eui64'"}
```

Reproduces identically on `2026.07.30-1` and on `main` — long-standing, not a regression.

### Why `422` is the right status

These are *semantic* rejections — "this combination of otherwise-valid fields does not make sense". The status is fine. The defect is only that the body shape is undeclared.

### Fix

PR to follow. It widens the **document** rather than rewriting the **bodies**, in one place, with zero behaviour change:

```python
# create_app(), wrapping FastAPI's own openapi()
props["detail"] = {"anyOf": [detail, {"type": "string"}], "title": "Detail"}
```

The string form is the established contract — the frontend's `formatApiError` already normalises string, array and object shapes (#31 / #186), so both render correctly today. Rewriting 270 response bodies to chase the schema would break clients outside this repo in order to fix a documentation defect.

Two alternatives were costed and rejected: reshaping the bodies into `ValidationError[]` breaks your own test suite (65 test files touch `detail`, many asserting `"…" in r.json()["detail"]`, which requires a string) and the UI; moving hand-raised rejects to `400` is 270 sites, the same breakage, and a client-visible status change on a long-stable API.

### Known remaining edge, not covered

Exactly one 422 raises a **dict** detail — `dns/router.py:4403`, `{"code": "DYNAMIC_UPDATE_UNSUPPORTED", "message": …}`. It is not reachable by the fuzz (both path parameters on `/dns/groups/{group_id}/zones/{zone_id}/update-acl` are `format: uuid`, so generated UUIDs 404 at `_require_zone` first), and the `anyOf` above does not admit it. Worth a follow-up; it is not one of the 29 rows.

### Also found in the same pass, separate and minor

`GET /api/v1/audit/export.pdf` and `GET /api/v1/conformity/export.pdf` fail with `UndefinedContentType` — the served media type is not in the declared `content` map. Happy to file separately.

### Not claimed here

The same pass reports ~580 "undocumented HTTP status code" results, almost all 404s from random UUIDs against `/{id}` paths that do not declare 404. That is a documentation gap rather than a defect and is deliberately not run per-commit. Raise separately if wanted.

### Environment

SpatiumDDI `2026.07.30-1` and `main` @ `eab2547`, single-node nested QA appliances, schemathesis 4.24.3.

amoona6:
Fixes #834

Root-caused and validated live on nested single-node QA appliances, across two releases and two rigs.

## 1. The document declared one 422 shape; the API returns two

FastAPI auto-declares `HTTPValidationError` (`detail`: array of `ValidationError`) on every operation that validates a parameter — honest for the 422s its own request validation produces. But **270 call sites** across `backend/app/api/v1` (57 router files, led by `ipam/router.py` 24, `dns/router.py` 22, `cutover/router.py` 18) raise `HTTPException(status_code=422, detail="…")` by hand for semantic validation a signature cannot express, and FastAPI serialises those as `{"detail": "<string>"}`.

The document said array, the wire said string, and nothing reconciled them — so every hand-raised 422 violated the API's own published contract.

Found by the per-commit conformance fuzz, which validates live responses against this very document:

```
GET /api/v1/ipam/addresses/search?limit=128&sort=null
-> 422 {"detail":"sort must be one of: address, description, hostname, last_seen, mac, status"}

  "sort must be one of: …" is not of type "array"
  Validated against the response schema for status code 422.
  Schema at /components/schemas/HTTPValidationError/properties/detail
```

Grouping every violation instance in a full pass over the affected set:

```
54  Validated against the response schema for status code 422
54  Schema at /components/schemas/HTTPValidationError
```

Unanimous — one cause, no exceptions. **29 of 39 red conformance rows on one build were this single defect.**

Widening the **document** rather than rewriting the **bodies** is deliberate. The string form is the established contract: the frontend's `formatApiError` already normalises string, array and object shapes (#31 / #186), so both render correctly today. Rewriting 270 response bodies to chase the schema would break clients outside this repo in order to fix what is a documentation defect.

The wrapper goes around FastAPI's own `openapi()` rather than re-deriving the document with `get_openapi`, so every generation setting the app carries (webhooks, separate input/output schemas, servers) keeps applying. That call caches into `app.openapi_schema` and returns the cached object, so this patches in place — the `anyOf` guard keeps it idempotent.

## 2. `page` was unbounded, so a deep offset overflowed the bigint

`page_size` has been bounded since #455 so "fetch all" could not sneak back in through an enormous page. `page` never was — and the offset is `(page - 1) * page_size`, while SQL `OFFSET` is a bigint. Every list endpoint was one query parameter away from a 500.

Measured on a live appliance at `DEFAULT_PAGE_SIZE = 100`:

| `page` | offset | status |
|---|---|---|
| 92233720368547758 | 9223372036854775700 | 200 |
| 92233720368547759 | 9223372036854775800 | 200 |
| 92233720368547760 | 9223372036854775900 | **500** |
| 92233720368547761 | 9223372036854776000 | **500** |

The boundary is exactly `2**63-1`, so the mechanism is the OFFSET overflow and nothing else.

Rig vmid 108 on `optiplex3`, `role=patch`, `suites=api`, against `dev-eab2547` as the baseline.

| | before (`dev-eab2547`) | after (`dev-7e75843`) |
|---|---|---|
| api rows | 827 | 1116 |
| fail | 40 | **12** |
| pass | 787 | 986 |
| `JsonSchemaError` rows | **29** | **0** |
| `DELETE /api/v1/dns-threat/mutes/{client_ip}` | **fail** | **pass** |

33 checks went from fail to pass.

**The lane budget alone would not have proven this.** These operations fail only ~50% of runs at `api_fuzz_examples: 3` and 100% at 25, so the previously-failing set was re-run on the fixed rig at 25 — the budget where the unfixed code fails every time:

```
12 failed, 82 passed, 1016 deselected in 601.09s

Validated against the response schema for status code …   0
Schema at /components/schemas/…                           0
```

Zero schema-conformance violations. All 12 residuals are the `502` gateway wave under fuzz load (one surfacing as a `hypothesis FlakyFailure` on a transient 502), unrelated to this change.

**No regression.** Four checks flipped pass→fail, all `500`s — `DELETE /settings/audit-forward-targets/{target_id}`, `POST /new-devices/allowlist`, `POST /new-devices/block`, `POST /overlays`. Re-running them on an **unfixed** rig (vmid 102, `2026.07.30-1`) reproduces the same 500s (`5 failed, 18 passed`), so they are pre-existing defects surfaced by different random sampling. Commit 1 changes only the generated document and cannot produce a runtime 500; commit 2 only *adds* an upper bound.

Six checks still fail from before and are untouched here: `schema-health` 500, two `export.pdf` `UndefinedContentType`, three import-preview 502s.

## Tests

`backend/tests/test_openapi_conformance.py` (new) — generated from the route table alone, no DB, no client:

- `test_validation_detail_allows_the_string_form_we_actually_return`
- `test_every_page_parameter_declares_an_upper_bound` — asserted over the document rather than the source, so it also covers list endpoints that build their own offset instead of calling `paginate()`, which is most of them.

These are regression guards for defects the conformance fuzz found on live appliances but nothing in CI could see: the fuzz validates real responses against this document, so a document that lies fails dozens of unrelated endpoints at once and reads as flake.

## Tradeoffs and known remaining work

- Widening `HTTPValidationError.detail` to a union makes a well-known FastAPI model loosely typed. The alternative — keeping it intact and making each 422 response `anyOf: [HTTPValidationError, ErrorDetail]` — generates better clients but rewrites the 422 of ~1086 operations. Happy to switch if you prefer that shape.
- **One uncovered edge:** exactly one 422 raises a dict detail (`dns/router.py:4403`, `DYNAMIC_UPDATE_UNSUPPORTED`), which the `anyOf` does not admit. It is not fuzz-reachable — both path params on that route are `format: uuid`, so generated UUIDs 404 at `_require_zone` first — so it is a latent gap rather than a live failure.
- The two `export.pdf` `UndefinedContentType` failures are a separate defect and are not addressed here.

### DIFF
--- a/.github/workflows/agent-e2e.yml
+++ b/.github/workflows/agent-e2e.yml
@@ -70,7 +70,7 @@ jobs:
       - uses: actions/checkout@v7
 
       - name: Install Helm
-        uses: azure/setup-helm@v4
+        uses: azure/setup-helm@v5
         with:
           version: v3.20.2
 
--- a/.github/workflows/build-appliance-builder.yml
+++ b/.github/workflows/build-appliance-builder.yml
@@ -30,18 +30,18 @@ jobs:
     steps:
       - uses: actions/checkout@v7
 
-      - uses: docker/setup-buildx-action@v3
+      - uses: docker/setup-buildx-action@v4
 
       - name: Log in to ghcr.io
         if: github.event_name != 'pull_request'
-        uses: docker/login-action@v3
+        uses: docker/login-action@v4
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Build (PR) / Build + push (main)
-        uses: docker/build-push-action@v6
+        uses: docker/build-push-action@v7
         with:
           context: appliance/builder
           push: ${{ github.event_name != 'pull_request' }}
--- a/.github/workflows/build-appliance.yml
+++ b/.github/workflows/build-appliance.yml
@@ -85,7 +85,7 @@ jobs:
     steps:
       - uses: actions/checkout@v7
       - name: Log in to GHCR
-        uses: docker/login-action@v3
+        uses: docker/login-action@v4
         with:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
@@ -119,7 +119,7 @@ jobs:
         # host: baltocdn.com" — Azure runner network policy).
         # bake-chart.sh shells out to ``helm package`` so helm must be
         # on PATH before the chart-bake step.
-        uses: azure/setup-helm@v4
+        uses: azure/setup-helm@v5
         with:
           version: v3.20.2
       - name: Fetch k3s binary + airgap tarball
--- a/.github/workflows/build-dhcp-images.yml
+++ b/.github/workflows/build-dhcp-images.yml
@@ -39,10 +39,10 @@ jobs:
     steps:
       - uses: actions/checkout@v7
 
-      - uses: docker/setup-qemu-action@v3
-      - uses: docker/setup-buildx-action@v3
+      - uses: docker/setup-qemu-action@v4
+      - uses: docker/setup-buildx-action@v4
 
-      - uses: docker/login-action@v3
+      - uses: docker/login-action@v4
         if: github.event_name != 'pull_request'
         with:
           registry: ghcr.io
@@ -51,7 +51,7 @@ jobs:
 
       - name: Image metadata
         id: meta
-        uses: docker/metadata-action@v5
+        uses: docker/metadata-action@v6
         with:
           images: ghcr.io/spatiumddi/dhcp-${{ matrix.flavor }}
           tags: |
@@ -61,7 +61,7 @@ jobs:
 
       - name: Build (PR — single-arch, load locally for Trivy)
         if: github.event_name == 'pull_request'
-        uses: docker/build-push-action@v5
+        uses: docker/build-push-action@v7
         with:
           context: .
           file: agent/dhcp/images/${{ matrix.flavor }}/Dockerfile
@@ -78,7 +78,7 @@ jobs:
 
       - name: Build and push (main / tag — multi-arch, push to ghcr)
         if: github.event_name != 'pull_request'
-        uses: docker/build-push-action@v5
+        uses: docker/build-push-action@v7
         with:
           context: .
           file: agent/dhcp/images/${{ matrix.flavor }}/Dockerfile
--- a/.github/workflows/build-dns-images.yml
+++ b/.github/workflows/build-dns-images.yml
@@ -42,10 +42,10 @@ jobs:
     steps:
       - uses: actions/checkout@v7
 
-      - uses: docker/setup-qemu-action@v3
-      - uses: docker/setup-buildx-action@v3
+      - uses: docker/setup-qemu-action@v4
+      - uses: docker/setup-buildx-action@v4
 
-      - uses: docker/login-action@v3
+      - uses: docker/login-action@v4
         if: github.event_name != 'pull_request'
         with:
           registry: ghcr.io
@@ -54,7 +54,7 @@ jobs:
 
       - name: Image metadata
         id: meta
-        uses: docker/metadata-action@v5
+        uses: docker/metadata-action@v6
         with:
           images: ghcr.io/spatiumddi/dns-${{ matrix.flavor }}
           tags: |
@@ -64,7 +64,7 @@ jobs:
 
       - name: Build (PR — single-arch, load locally for Trivy)
         if: github.event_name == 'pull_request'
-        uses: docker/build-push-action@v5
+        uses: docker/build-push-action@v7
         with:
           context: .
           file: agent/dns/images/${{ matrix.flavor }}/Dockerfile
@@ -82,7 +82,7 @@ jobs:
 
       - name: Build and push (main / tag — multi-arch, push to ghcr)
         if: github.event_name != 'pull_request'
-        uses: docker/build-push-action@v5
+        uses: docker/build-push-action@v7
         with:
           context: .
           file: agent/dns/images/${{ matrix.flavor }}/Dockerfile
--- a/.github/workflows/build-looking-glass-images.yml
+++ b/.github/workflows/build-looking-glass-images.yml
@@ -39,10 +39,10 @@ jobs:
     steps:
       - uses: actions/checkout@v7
 
-      - uses: docker/setup-qemu-action@v3
-      - uses: docker/setup-buildx-action@v3
+      - uses: docker/setup-qemu-action@v4
+      - uses: docker/setup-buildx-action@v4
 
-      - uses: docker/login-action@v3
+      - uses: docker/login-action@v4
         if: github.event_name != 'pull_request'
         with:
           registry: ghcr.io
@@ -51,7 +51,7 @@ jobs:
 
       - name: Image metadata
         id: meta
-        uses: docker/metadata-action@v5
+        uses: docker/metadata-action@v6
         with:
           images: ghcr.io/spatiumddi/looking-glass
           tags: |
@@ -61,7 +61,7 @@ jobs:
 
       - name: Build (PR — single-arch, load locally for Trivy)
         if: github.event_name == 'pull_request'
-        uses: docker/build-push-action@v5
+        uses: docker/build-push-action@v7
         with:
           context: .
           file: agent/looking-glass/images/${{ matrix.flavor }}/Dockerfile
@@ -78,7 +78,7 @@ jobs:
 
       - name: Build and push (main / tag — multi-arch, push to ghcr)
         if: github.event_name != 'pull_request'
-        uses: docker/build-push-action@v5
+        uses: docker/build-push-action@v7
         with:
           context: .
           file: agent/looking-glass/images/${{ matrix.flavor }}/Dockerfile
--- a/.github/workflows/build-supervisor-image.yml
+++ b/.github/workflows/build-supervisor-image.yml
@@ -35,10 +35,10 @@ jobs:
     steps:
       - uses: actions/checkout@v7
 
-      - uses: docker/setup-qemu-action@v3
-      - uses: docker/setup-buildx-action@v3
+      - uses: docker/setup-qemu-action@v4
+      - uses: docker/setup-buildx-action@v4
 
-      - uses: docker/login-action@v3
+      - uses: docker/login-action@v4
         if: github.event_name != 'pull_request'
         with:
           registry: ghcr.io
@@ -47,7 +47,7 @@ jobs:
 
       - name: Image metadata
         id: meta
-        uses: docker/metadata-action@v5
+        uses: docker/metadata-action@v6
         with:
           images: ghcr.io/spatiumddi/spatium-supervisor
           tags: |
@@ -66,7 +66,7 @@ jobs:
 
       - name: Build (PR — single-arch, load locally for Trivy)
         if: github.event_name == 'pull_request'
-        uses: docker/build-push-action@v5
+        uses: docker/build-push-action@v7
         with:
           context: .
           file: agent/supervisor/images/supervisor/Dockerfile
@@ -82,7 +82,7 @@ jobs:
 
       - name: Build and push (main / tag — multi-arch, push to ghcr)
         if: github.event_name != 'pull_request'
-        uses: docker/build-push-action@v5
+        uses: docker/build-push-action@v7
         with:
           context: .
           file: agent/supervisor/images/supervisor/Dockerfile
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -177,9 +177,9 @@ jobs:
         include: ${{ fromJSON(needs.gate.outputs.images) }}
     steps:
       - uses: actions/checkout@v7
-      - uses: docker/setup-qemu-action@v3
-      - uses: docker/setup-buildx-action@v3
-      - uses: docker/login-action@v3
+      - uses: docker/setup-qemu-action@v4
+      - uses: docker/setup-buildx-action@v4
+      - uses: docker/login-action@v4
         if: github.event.inputs.dry_run != 'true'
         with:
           registry: ${{ env.REGISTRY }}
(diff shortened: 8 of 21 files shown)


amoona6:
I built the same fix independently in a parallel QA session before spotting this PR, and landed on a different variant of the same document-side strategy. **Not opening a competing PR** — the approach here is right and the diff is smaller. Two deltas from my version seem worth folding in, one of them a correctness gap this PR's own issue already flags.

### 1. The `anyOf` does not admit the object-shaped `detail`

`#834` lists the dict-shaped 422 as a known remaining gap. It is a single site, and it is reachable by a real client:

```python
# backend/app/api/v1/dns/router.py:4402-4405
raise HTTPException(
    status_code=422,
    detail={"code": "DYNAMIC_UPDATE_UNSUPPORTED", "message": str(exc)},
) from exc
```

That fires when a DNS driver rejects a zone's dynamic-update ACL — a normal operator path, not an edge case. With `anyOf: [<array>, {"type": "string"}]` its body still violates the declared schema, exactly as before this PR.

It will never be caught by a test, either: the conformance fuzz generates random UUIDs for `{zone_id}`, so it 404s long before reaching the driver call. The document is simply wrong for it, silently and permanently.

One extra member closes it:

```python
props["detail"] = {
    "anyOf": [detail, {"type": "string"}, {"type": "object"}],
    "title": detail.get("title", "Detail"),
}
```

I verified the site count independently: `git grep -c 'status_code=422' -- backend/app/api/v1` gives **270**, matching the commit message, and exactly **one** of those passes a dict.

### 2. Widening `HTTPValidationError` in place vs. adding a sibling

This is a judgement call rather than a bug, so take it or leave it.

`HTTPValidationError` is FastAPI's well-known model and client generators recognise it by name. Widening its `detail` in place means the model itself becomes loosely typed everywhere it is referenced — so the type a generator emits for *the framework's own validation error* is now `array | string`, including in the many operations that can only ever return the array form. (I can't give an exact split: hand-raised rejects come from shared helpers like `_validate_sort_order` and the pagination guards, so which operations can reach one isn't statically obvious — which is also why my variant applies its union uniformly rather than to a curated list.)

The alternative I used keeps that model untouched and adds a sibling, switching each response to a union of two concrete shapes:

```python
components.setdefault("ErrorDetail", {
    "title": "ErrorDetail",
    "type": "object",
    "properties": {"detail": {"anyOf": [{"type": "string"}, {"type": "object"}], "title": "Detail"}},
    "required": ["detail"],
})
# then, per 422 response whose schema is exactly $ref HTTPValidationError:
media["schema"] = {"anyOf": [
    {"$ref": "#/components/schemas/HTTPValidationError"},
    {"$ref": "#/components/schemas/ErrorDetail"},
]}
```

Trade-off, stated plainly: this rewrites 971 response objects instead of one component, so it is a bigger post-process and a longer document. It buys an unchanged `HTTPValidationError` and a named, documented type for the hand-raised shape. The current approach is defensible if you would rather keep the diff minimal — in which case delta 1 still applies on its own.

### Validation of the variant

Measured on the live generated document (1110 operations): 971 declare a 422, all 971 carry the union, 0 remain bare, `HTTPValidationError.detail` stays `type: array`, and the result still passes `openapi-spec-validator` as OpenAPI 3.1.0.

On a nested QA rig built from the variant, the full conformance walk went from **29 `JsonSchemaError` operations to 0**. Re-running the ipam slice (118 operations, which covers 12 of the 29) at 25 examples — the budget at which these fail deterministically — gives **0 schema violations**; all 16 remaining failures are `5xx` assertions, i.e. the separate gateway/500 defect class, not contract violations.

Being precise about one of them: `GET /api/v1/ipam/addresses/search` is still red on that rig, but now as a 5xx rather than a `JsonSchemaError`. A document-only change cannot alter a runtime response, and this rig is a single-node 8 GiB patch box where the 5xx wave is broader than on the 3-node flagship — but I did not run an unpatched control of the same rig class, so I'm reporting that as observed rather than as proven-unrelated.

Backend suite: 3700 passed / 1 failed; that failure (`test_heartbeat_wake_shortens_the_hold`) reproduces on unmodified `main`, so it is pre-existing.

Finally, a caveat that applies to **any** validation of this fix, including this PR's: at the essential lane's `api_fuzz_examples: 3` these operations only fail about **50%** of runs, so a green run at that budget proves very little. Measured on one rig, one code state, 20 runs with a fresh hypothesis example DB each time: 5/10 failures at 3 examples, **10/10 at 25**. Re-running the previously-failing set at 25 is what actually demonstrates the fix.

amoona6:
Adding the **unpatched control run** the previous comment flagged as missing, plus three corrections to the numbers in the PR description. Same fix, independently validated on a different rig class; nothing here changes the approach, which I agree is right.

## 1. The unpatched control, run

> *"I did not run an unpatched control of the same rig class, so I'm reporting that as observed rather than as proven-unrelated."*

I ran it. Every check that flipped pass→fail on the patched build was re-run against an **unpatched appliance** (`2026.07.30-1`, containing neither commit), and **every one reproduces there**.

The full essential-lane eval on the final sha showed 9 pass→fail — more than the 4 in the PR description. All 9:

| Endpoint | unpatched build | note |
|---|---|---|
| `POST /api/v1/new-devices/allowlist` | **500** | |
| `POST /api/v1/new-devices/block` | **500** | unvalidated `str` → `MACADDR` column, see below |
| `POST /api/v1/overlays` | **500** | |
| `GET /api/v1/overlays` | **500** | |
| `POST /api/v1/appliance/tls/csr` | `JsonSchemaError` | |
| `POST /api/v1/customers` | **500** | hypothesis `FlakyFailure` |
| `POST /api/v1/domains` | **500** | hypothesis `FlakyFailure` |
| `POST /api/v1/vrfs` | **500** | hypothesis `FlakyFailure` |
| `PUT /api/v1/dnsbl/settings` | **500** | |
| `DELETE /api/v1/settings/audit-forward-targets/{target_id}` | passes at 200 examples | one-off draw |

Three carry hypothesis's own verdict — *"Failed on the first call but did not on a subsequent one"*. So the no-regression claim is now **proven**, not inferred from "a document-only change cannot alter a runtime response".

The same control surfaced 7 pre-existing 500s not previously recorded anywhere: `GET /customers`, `POST /dnsbl/lists`, `GET /dnsbl/listings`, `GET /dnsbl/listings/by-ip/{ip}`, `POST /dnsbl/pinned`, `GET /domains`, `POST /multicast/domains`, plus `POST /api/v1/dns/blocklists`.

**Why the noise floor makes this control mandatory.** Two runs of the *identical* build, no code change between them, disagreed on **7 checks**:

```
DELETE /settings/audit-forward-targets/{target_id}   fail -> pass
GET    /appliance/appliances/{id}/dependents         pass -> fail
GET    /dhcp/import/windows/servers                  pass -> fail
POST   /dhcp/agents/register                         pass -> fail
POST   /dns/agents/register                          pass -> fail
POST   /ipam/spaces                                  pass -> fail
POST   /overlays                                     fail -> pass
```

A regression set of 4–9 is indistinguishable from that noise without a control.

## 2. Three corrections to the PR description

**a. "All 12 residuals are the `502` gateway wave" is not right.** Grouping the 12 by cause:

```
500 × 7   appliance/appliances/{id} ×2, settings/audit-forward-targets,
          admin/postgres/schema-health, new-devices/allowlist,
          new-devices/block, overlays
502 × 3   dns/import/powerdns, dns/import/technitium, ipam/import/netbox  (previews)
UndefinedContentType × 2   audit/export.pdf, conformity/export.pdf
```

Only 3 of 12 are 502s. This also contradicts the description's own later paragraph, which lists them correctly.

**b. The row-count table compares two different denominators.** "api rows | 827 | 1116" is *pass+fail* before against *total including skips* after:

| | `dev-eab2547` | `dev-7e75843` |
|---|---|---|
| total rows | 1195 | 1116 |
| pass + fail | 827 | 998 |
| skips | 368 | 118 |

**c. The baseline is not like-for-like, which explains the skip gap.** The eval lane runs `api_tier: deep` (`classes.yaml`); a `role=patch` devbuild sets `api_tier: core` (`rig_up_prox.py`), deselecting the `test_deep_*` tier. Different test sets, so total row counts are not comparable — compare `pass+fail`, or hold the tier constant.

None of these affect the fix or the `JsonSchemaError` 29 → 0 result.

## 3. Standardized lane verdict

The essential-lane eval on the final sha, for completeness:

| | baseline `dev-eab2547` | fixed |
|---|---|---|
| verdict | BLOCK | BLOCK |
| error anomalies | 40 | **27** |
| pass rate | 68.9% | **88.5%** |
| `JsonSchemaError` | **29** | **0** |

Still BLOCK because the `patch` class blocks on *any* error anomaly and the tree carries dozens unrelated to this change — no scoped fix can reach PASS. Worth knowing before reading the verdict as a signal about this PR.

## 4. Separate defect found while controlling the above

`POST /api/v1/new-devices/block` 500s on any non-MAC string: `BlockBody.mac_address` is a bare `str`, but the column is PostgreSQL `MACADDR` (`models/dhcp.py`, `DHCPMACBlock.mac_address`), so the value reaches a `macaddr` cast unvalidated. Pre-existing, unrelated to this PR, happy to file separately.

## 5. Agreement on the fuzz-budget caveat

Independently confirmed: at `api_fuzz_examples: 3` this check passes and fails at roughly even odds. Its residual failures on the patched build are the 502 wave — at 200 examples hypothesis diagnosed one itself:

```
hypothesis.errors.FlakyFailure: produces unreliable results:
  Failed on the first call but did not on a subsequent one
Failed to reproduce exception. Expected:
  AssertionError: 5xx from GET /api/v1/new-devices/sightings: 502
```

Across runs the 502s hit a shifting set of 41–66 endpoints, only 17 shared between two runs. It is the gateway under fuzz load, not any handler — so the originally-reported cell will keep flickering after this merges, for a cause this PR correctly does not claim to fix.

Issue #707, opened 2026-07-27
Imported DNS records are rendered to the BIND zone file but are not applied to the running dynamic zone


Cmonnich:
### SpatiumDDI version

v2026.07.21-1

### Deployment

OS appliance

### Area

DNS

### Actual behavior

The imported records are stored and rendered to disk, but the running BIND zone remains unchanged until each record is edited manually.

### Current workaround

For now, the reliable choices are:

1. Create records individually through the GUI.
2. Import them, then open and save each record to trigger the update operation.

### Expected behavior

After a successful import, Spatium should either:

- enqueue record operations for every created or updated record, or

- safely reload/thaw the complete dynamic zone after replacing its master file.

### Reproduction steps

1.  Create a Primary Forward zone assigned to a BIND server.
2.  Import a valid BIND zone file containing multiple A records.
3.  Confirm the records appear in the SpatiumDDI GUI.
4.  Confirm the records appear in: /var/lib/spatium-dns-agent/rendered/zones/<zone>.db
5. Query an imported record directly against the BIND server.
6. Observe NXDOMAIN.
7. Edit one imported record in the GUI and save it.
8. Observe record_op_applied in the DNS-agent log.
9. Query the record again and observe that it now resolves.

### Relevant logs / tracebacks

```shell

```

### Pre-submission checks

- [x] I searched existing issues and this is not a duplicate
- [x] I am running a recent SpatiumDDI release (within the last ~2 releases)

mzac:
@Cmonnich Thanks for the bug report, I'll look into it as soon as I can.

amoona6:
Reopening request — this reproduces on 2026.07.30-1 and on current `main`, by a
different route than the import path that was fixed.

The original report offered two remedies:

> - enqueue record operations for every created or updated record, **or**
> - safely reload/thaw the complete dynamic zone after replacing its master file.

The first was implemented. Split-horizon (#24) then removed it: for a DNS server
group that has **any view**, the control plane deliberately stops dispatching
record ops — an nsupdate to loopback cannot target a specific view — retires the
queued ops as `applied`, and propagates the change by re-rendering the zone file
instead (`backend/app/services/dns/agent_config.py`, the `has_views` branch;
records are folded into `structural_etag` at the same site so the agent does a
full re-render).

The second remedy was never implemented, so on any group with a view the
original symptom is back in full: **records are rendered to the BIND zone file
and never applied to the running dynamic zone.** It is not limited to imports
[...]

Issue #704, opened 2026-07-25
dns-agent: BIND9 agent starts named twice — SO_REUSEPORT hides it, rndc becomes non-deterministic (query logging silently dead)


mzac:
Found while testing #699 end-to-end against a real BIND9 agent on the dev stack.

## Symptom

Enabling **query logging** on a DNS server group does nothing on a running BIND9 agent. The config renders correctly, `/var/log/named/queries.log` is created, named reconfigures — and not a single query is logged. `rndc status` reports:

```
query logging is OFF
```

Restarting the container with the **byte-identical** config brings it up `query logging is ON`.

Worse, `rndc querylog on` returns rc=0 and changes nothing — so the obvious workaround also appears broken.

## Root cause: the agent starts `named` twice

```
{"pid": 14, "event": "named_started", ... "timestamp": "2026-07-25T20:06:19.126018Z"}
{"pid": 32, "event": "named_started", ... "timestamp": "2026-07-25T20:06:19.244228Z"}
```

Two `named` processes, same parent, same config, 118 ms apart. They don't fail loudly because BIND uses `SO_REUSEPORT` — **both** bind `:53` and `:953`, and the kernel load-balances between them.

That makes `rndc` non-deterministic: `rndc querylog on` flips the flag on whichever instance answered, while the next `rndc status` — and the actual DNS queries — may land on the other. Query logging is per-process state, so it looks like the toggle silently fails.

`Bind9Driver.daemon_running()` trusts `self.daemon_pid`, which is **per-instance** state:

```python
def daemon_running(self) -> bool:
    if self.daemon_pid is None:
        return False          # ← a second driver object always says "not running"
```

`start_daemon()` is called from two places (`supervisor.py:49` and the config-apply path in `bind9.py`). Any second driver object — or any race between those paths — sees `daemon_running() is False` and spawns another daemon.

## Blast radius

Wider than query logging, because everything `rndc`-driven inherits the coin flip:

- **Query logging** appears permanently off → Logs → DNS Queries is empty, and so is anything built on it (this is how it surfaced: #699's DNS threat analytics had nothing to score, and an operator would reasonably conclude the detector is broken rather than that the log feed is).
- **`rndc reload` / `reconfig` / `notify`** reach one instance; the other keeps serving stale zone state.
- **DNSSEC operations** (`rndc dnssec -rollover`, signing status reads) may act on one instance and be read back from the other.
- Two full resolver instances also means double the cache memory for no benefit.

[...]

PR #842, opened 2026-08-07
fix(dns-agent): reload the zone files `rndc reconfig` ignores, so a record reaches the wire

### LINKED ISSUE #838: dns-agent: BIND9 agent sends `rndc reconfig`, never reloads changed zone files — views-mode groups never serve a record created after first load
### SpatiumDDI version

2026.07.30-1 (also reproduced on current `main`, merge-base `f4882c31`)

### Deployment

OS appliance (nested 3-node QA cluster, BIND9 agent)

### Area

DNS

### What happened?

**In views mode, a DNS group has no working record-delivery path at all.** A record created through the API is accepted (`201`), written to the DB, rendered into the zone file on the agent — and never served. Not late: never. The running `named` keeps answering from whatever it loaded at boot, indefinitely.

This is adjacent to #707 but is **not** the same defect, and #711 does not fix it. #711 made the *importer* enqueue `DNSRecordOp` rows so imported records reach the daemon over RFC 2136. That route does not exist for a group with views: `build_config_bundle` deliberately disables it and says so —

https://github.com/spatiumddi/spatiumddi/blob/main/backend/app/services/dns/agent_config.py#L295-L302

```python
elif has_views:
    # Split-horizon: the incremental RFC 2136 path can't target a
    # specific view (an nsupdate to loopback lands in whichever view
    # matches 127.0.0.1, not necessarily the record's view). Records
    # are folded into the structural fingerprint below so every record
    # change triggers a full, view-correct re-render instead. Retire any
    # queued ops as ``applied`` — the bundle the agent is about to render
    # already reflects them — so they don't pile up in ``pending``.
```

So in views mode the *entire* contract rests on "re-render the zone file, then reload the daemon". The re-render happens correctly. The reload does not.

### Root cause

`Bind9Driver.swap_and_reload()` signals the daemon with **`rndc reconfig`**:

https://github.com/spatiumddi/spatiumddi/blob/main/agent/dns/spatium_dns_agent/drivers/bind9.py#L1070

```python
cmd.append("reconfig")
```

`rndc reconfig` re-reads `named.conf` and loads zones that are **new to the configuration**. By design it does *not* re-read the file of a zone that already exists. Every change to an existing zone's *contents* is therefore invisible to the running daemon — permanently.

The failure is silent end to end. The agent renders, atomically swaps, logs `structural_reload_applied`, and moves on; the control plane retires the queued op as `applied` on the way past (per the comment above). Nothing anywhere reports an error.

Consequence chain for a group with views:

1. `POST /dns/groups/{g}/zones/{z}/records` → `201`, `DNSRecord` row written.
2. Bundle rotates `structural_etag` (records are folded in, views mode).
3. Agent re-renders the zone file — correctly, record present, SOA serial bumped.
4. Agent runs `rndc reconfig` → **zone data untouched in memory**.
5. Wire keeps serving the boot-time serial forever.

Records that were present when the zone was *first* loaded do serve, which makes this look like "DNS works" on a fresh appliance and masks the bug until someone adds a record.

### Actual behavior

Measured live on a nested 3-node appliance cluster at 2026.07.30-1, zone `lab.ddipg.test` inside view `meridian-internal`:

```
rendered file  /var/lib/spatium-dns-agent/rendered/zones/meridian-internal/lab.ddipg.test.db
               @ IN SOA ns1.lab.ddipg.test. admin.lab.ddipg.test. ( 2026080612 3600 600 86400 300 )
               drill-672a7b 3600 IN A 192.0.2.77          ← the record IS on disk

wire           dig -b 10.10.10.254 @<appliance> lab.ddipg.test SOA
               lab.ddipg.test. 3600 IN SOA ... 2026080603 3600 600 86400 300
                                              ^^^^^^^^^^ pinned at the boot-time serial

               dig -b 10.10.10.254 @<appliance> drill-672a7b.lab.ddipg.test A
               []                                          ← after 91s
```

Nine serial bumps on disk, zero on the wire. Meanwhile seeded records from first load answer normally:

```
probe.lab.ddipg.test.  3600 IN A 10.10.50.10     ✓
canary.lab.ddipg.test. 3600 IN A 192.0.2.42      ✓
```

Agent log over the same window — the re-render fires every time, and there is no error and no `record_op_applied` anywhere:

```json
{"structural_etag":"sha256:001b0caf…","event":"structural_reload_applied","level":"info","timestamp":"2026-08-06T15:00:11.565534Z"}
{"structural_etag":"sha256:ecc69599…","event":"structural_reload_applied","level":"info","timestamp":"2026-08-06T15:01:27.989330Z"}
{"structural_etag":"sha256:cbb9c1b5…","event":"structural_reload_applied","level":"info","timestamp":"2026-08-06T15:03:02.645582Z"}
```

And the op rows are all retired as `applied` with nothing having reached the wire (expected, per the views-mode comment — noted here only so the DB state isn't mistaken for evidence that delivery worked):

```
delete|lab.ddipg.test.|applied|2026-08-06 15:10:10.288564+00
create|lab.ddipg.test.|applied|2026-08-06 15:09:38.231492+00
delete|lab.ddipg.test.|applied|2026-08-06 15:08:31.834542+00
create|lab.ddipg.test.|applied|2026-08-06 15:07:45.892300+00
```

### Which reload primitive actually works

Tested in order on the live appliance, immediately after a create, with the record on disk and absent from the wire:

| primitive | result |
|---|---|
| `rndc reconfig` (what the agent does today) | wire unchanged — serial still `2026080603`, `[]` |
| `rndc reload lab.ddipg.test` | **refused**: `rndc: 'reload' failed: dynamic zone` |
| `rndc reload` (server-wide) | `server reload successful` → record served immediately |
| `rndc freeze lab.ddipg.test IN meridian-internal` + `thaw` | `A zone reload and thaw was started.` → record served immediately |

Per-zone `reload` is not usable because zones render with `allow-update` (#641), so BIND treats them as dynamic. Server-wide `rndc reload` works in views mode because no journal exists there — RFC 2136 is disabled, so nothing has ever written one. `freeze`/`thaw` is the journal-safe generalization and is the primitive #707's reporter asked for ("safely reload/thaw the complete dynamic zone after replacing its master file").

### What did you expect?

A record accepted by the API should be answerable on the wire, in views mode as in any other. Concretely: `swap_and_reload()` should issue a signal that re-reads changed zone *files*, not only the configuration.

### Steps to reproduce

1. Install an OS appliance with a BIND9 DNS server group that has at least one **view** (split-horizon). Any group with views reproduces it; a group without views does not, which is why this survived #711.
2. Create a primary forward zone in that group and let the agent render it.
3. Query a record that existed at first load — it resolves.
4. `POST /api/v1/dns/groups/{group}/zones/{zone}/records` with a new A record. Observe `201`.
5. Confirm the record is present in `/var/lib/spatium-dns-agent/rendered/zones/<view>/<zone>.db` and that its SOA serial has advanced.
6. Query the new name against the appliance from an in-view vantage. Observe no answer — indefinitely.
7. Compare the zone's SOA serial on the wire to the one in the rendered file; the wire is pinned at the serial it loaded at boot.
8. Run `rndc reload` (or view-qualified `freeze`/`thaw`) inside the agent container. The record resolves immediately.

### Suggested fix

One line in `Bind9Driver.swap_and_reload()` — `reconfig` → `reload`. Proven live: the record went to the wire immediately. A PR with the change plus a regression test pinning the primitive follows this issue.

Worth considering beyond the minimal fix: `freeze`/`thaw` per changed zone (view-qualified) is the primitive that is also correct once a zone has a journal, which is the case for groups *without* views where RFC 2136 does write one. `rndc reload` is sufficient and proven for the views-mode path this issue is about; I did not test it against a dynamic zone carrying a live `.jnl`.

### Relevant logs / tracebacks

```shell
# after a create, inside the dns-bind9 container
$ rndc -c /var/lib/spatium-dns-agent/rndc.conf reload lab.ddipg.test
rndc: 'reload' failed: dynamic zone

$ rndc -c /var/lib/spatium-dns-agent/rndc.conf reload
server reload successful

$ rndc -c /var/lib/spatium-dns-agent/rndc.conf freeze lab.ddipg.test IN meridian-internal
$ rndc -c /var/lib/spatium-dns-agent/rndc.conf thaw  lab.ddipg.test IN meridian-internal
A zone reload and thaw was started.
Check the logs to see the result.
```

### Environment details

Nested 3-node appliance cluster (seed + 2 members) on Proxmox, BIND9 driver, group `meridian-internal` with one view, zones rendered with `allow-update { key "spatium-meridian-dns"; }`. Reproduced identically on release 2026.07.30-1 and on two builds of current `main`; the two automated checks that cover this path have never once passed since they were written.

### Pre-submission checks

- [x] I searched existing issues and this is not a duplicate — #707 is closed by #711, which fixes the import path via RFC 2136 and cannot apply to a views-mode group, where that path is disabled by design.
- [x] I am running a recent SpatiumDDI release (within the last ~2 releases)

amoona6:
Fixes #838 — in views mode a record accepted by the API never reaches the wire.

## What the issue asked for, and what this does instead

#838's "Suggested fix" proposed one line: `reconfig` → `reload`. That is proven
to work for the views-mode path the issue is about, but it is not sufficient in
general, and this PR does the thing the issue files under *worth considering*
instead. The reason is in the issue's own primitive table:

```
rndc reload <zone>   ->  refused: 'reload' failed: dynamic zone
```

Every primary zone this product renders is dynamic in named's eyes, because
`_render_allow_update` emits the group's loopback TSIG grant unconditionally —
including on zones with `dynamic_update_enabled: false`. named will not re-read
a dynamic zone's master file on a plain `reload`, since that would discard
journal contents. Server-wide `rndc reload` gets away with it *only* in views
mode, where RFC 2136 is disabled so no journal has ever been written. On a group
without views, the same call would be a data-loss hazard.

So `swap_and_reload` now keeps `reconfig` for what `reconfig` is actually for —
config changes and zones added or removed — and then does **freeze → reload →
thaw per rendered zone, scoped `in <view>` under split-horizon**. That is the
journal-safe generalization, and it is the primitive #707's reporter originally
asked for.

## Change

- `Bind9Driver.swap_and_reload()` issues the per-zone reload after `reconfig`.
- New `rendered_zone_views()` reads the zone/view layout back off the rendered
  tree rather than off the bundle, so it cannot drift from what was actually
  written to disk.
- Per-zone best-effort: one zone that will not reload does not strand the rest.

## Why the failure was invisible

Nothing reported an error anywhere along the chain. The agent rendered, swapped
atomically, logged `structural_reload_applied` and moved on; the control plane
retired the queued op as `applied` on the way past, which it is right to do in
views mode (`agent_config.py:295-315`). Records present at first load kept
serving, so a fresh appliance looks healthy until someone adds a record.

The change is now visible doing the work — a new log line pairs with every
structural reload:

```json
{"zones": 5, "event": "bind9_rendered_zones_reloaded", "timestamp": "..."}
{"structural_etag": "sha256:557cc70e…", "event": "structural_reload_applied", "timestamp": "..."}
```

Nested 3-node appliance cluster on a QA rig, group `meridian-internal` with one
view, zone `lab.ddipg.test`. Before-fix numbers are from a rig at the same ref;
after-fix from a rig built from this branch.

| | before | after |
|---|---|---|
| create → served on the wire | never (0/24 polls over 120 s) | **5 s**, then 6/6 digs |
| delete → withdrawn from the wire | never | **5 s** |
| targeted `suites/api/test_deep_wire.py -m integration` | fail | **3 passed** in 49 s |
| `test_deep_wire_dns_record_serves` | fail | pass |
| `test_deep_wire_dns_delete_stops_serving` | fail (precondition) | pass |

Both of those checks had **never passed since they were written** — 0 flips
across 146 tags, no flake signature — so this is not a lucky flip.

## Tests

`agent/dns/tests/test_zone_reload.py` — 6 new cases: the rendered-tree layout in
both shapes (views and flat), that `swap_and_reload` issues freeze/reload/thaw
per zone rather than only `reconfig`, view scoping of the rndc calls, and that
one unreloadable zone does not strand the others.

`agent/dns`: **183 passed, 8 skipped**. `ruff check` introduces no new findings
(16 on this branch, 16 on `main` — the two that land on touched files are
pre-existing `RUF100` unused-`noqa` on `daemon_version`).

## Not in this PR

- The duplicate-`named` defect that shares this code path is **#704**, which is
  closed but still reproduces. It has its own PR; the two are independent and
  touch disjoint regions of `bind9.py`.
- `_render_allow_update` emitting the loopback grant unconditionally is what
  forces the freeze/thaw dance at all. Narrowing it to
  `dynamic_update_enabled` zones would be a cleaner fix but changes the
  security surface, so it is left as a note.

### DIFF
--- a/agent/dns/spatium_dns_agent/drivers/bind9.py
+++ b/agent/dns/spatium_dns_agent/drivers/bind9.py
@@ -1063,23 +1063,99 @@ class Bind9Driver(DriverBase):
         # fall back to SIGHUP which named handles as a config + zone reload.
         rndc_ok = False
         if shutil.which("rndc"):
-            cmd = ["rndc"]
-            agent_conf = self.state_dir / "rndc.conf"
-            if agent_conf.exists():
-                cmd += ["-c", str(agent_conf)]
-            cmd.append("reconfig")
-            res = subprocess.run(cmd, capture_output=True, text=True, check=False)
+            base = self._rndc_base()
+            # ``reconfig`` picks up config changes and zones that were ADDED or
+            # REMOVED — but by BIND's documented definition it "does not reload
+            # existing zone files even if they have changed". That was fine
+            # while a record edit rode the RFC 2136 path. Under split-horizon
+            # (issue #24) it is not: the control plane deliberately stops
+            # dispatching record ops for a group with views — an nsupdate to
+            # loopback cannot target a view — and propagates record changes by
+            # RE-RENDERING the zone file instead
+            # (backend/app/services/dns/agent_config.py, has_views branch).
+            # ``reconfig`` never reads that file back, so on any group with a
+            # view every record created after the zone's first load was written
+            # to disk and never served: the API returned 201 and the wire
+            # answered NXDOMAIN forever. Proven live on a 3-node QA rig
+            # 2026-08-06 — the record was present in
+            # rendered/zones/<view>/<zone>.db while dig returned NXDOMAIN, and
+            # the zone's own ``rndc zonestatus`` still showed the serial and
+            # node count from the previous load.
+            res = subprocess.run(
+                [*base, "reconfig"], capture_output=True, text=True, check=False
+            )
             rndc_ok = res.returncode == 0
             if not rndc_ok:
                 log.warning(
                     "rndc_failed_falling_back_to_sighup", stderr=res.stderr.strip()
                 )
+            else:
+                self._reload_rendered_zones(base)
         if not rndc_ok and self.daemon_pid:
             try:
                 os.kill(self.daemon_pid, signal.SIGHUP)
                 log.info("named_sighup_sent", pid=self.daemon_pid)
             except OSError as e:
                 log.error("named_sighup_failed", error=str(e))
+
+    def rendered_zone_views(self) -> list[tuple[str, str | None]]:
+        """``(zone_name, view_name)`` for every zone file we just rendered.
+
+        Read back off the rendered tree rather than threaded down from
+        ``render()`` so it cannot drift from what is actually on disk — the
+        layout is ``rendered/zones/<view>/<zone>.db`` under split-horizon and
+        ``rendered/zones/<zone>.db`` without views, which is exactly the
+        ``file_prefix`` contract ``_zone_stanza`` writes.
+        """
+        root = self.state_dir / self.rendered_dir_name / "zones"
+        out: list[tuple[str, str | None]] = []
+        try:
+            entries = sorted(root.iterdir())
+        except OSError:
+            return out
+        for entry in entries:
+            if entry.is_dir():
+                for zf in sorted(entry.glob("*.db")):
+                    out.append((zf.name[: -len(".db")], entry.name))
+            elif entry.name.endswith(".db"):
+                out.append((entry.name[: -len(".db")], None))
+        return out
+
+    def _reload_rendered_zones(self, base: list[str]) -> None:
+        """Make named re-read the zone files ``reconfig`` just ignored.
+
+        Every primary zone we render carries an ``allow-update`` clause — the
+        group's loopback TSIG grant is ALWAYS included so control-plane record
+        ops can flow (see ``_render_allow_update``), which makes the zone
+        DYNAMIC as far as named is concerned. named will not re-read a dynamic
+        zone's file on a plain reload, because doing so would silently discard
+        journal contents; it has to be frozen first. So: freeze → reload →
+        thaw, per zone, per view. ``freeze``/``thaw`` fail harmlessly on a zone
+        that is not dynamic, and the plain ``reload`` covers that case, so one
+        sequence is correct for both.
+
+        Best-effort by design: a zone that will not reload must not stop the
+        rest from reloading, and it is already reported through the daemon's
+        own status channel.
+        """
+        for zname, view in self.rendered_zone_views():
+            scope = [zname] + (["in", view] if view else [])
+            for verb in ("freeze", "reload", "thaw"):
+                res = subprocess.run(
+                    [*base, verb, *scope],
+                    capture_output=True,
+                    text=True,
+                    check=False,
+                )
+                if res.returncode != 0 and verb == "reload":
+                    log.warning(
+                        "bind9_zone_reload_failed",
+                        zone=zname,
+                        view=view,
+                        stderr=res.stderr.strip()[:200],
+                    )
+        log.info("bind9_rendered_zones_reloaded",
+                 zones=len(self.rendered_zone_views()))
 
     # ── Record ops (RFC 2136 over loopback) ─────────────────────────────────
 
--- a/agent/dns/tests/test_zone_reload.py
+++ b/agent/dns/tests/test_zone_reload.py
@@ -1,0 +1,128 @@
+"""A rendered zone file is worthless if named never reads it back (#704).
+
+``rndc reconfig`` — what ``swap_and_reload`` used to issue on its own — is
+defined by BIND as "reload the configuration file and load new zones, but do
+not reload existing zone files even if they have changed". That is the right
+primitive only while a record edit reaches the daemon some OTHER way. Under
+split-horizon (issue #24) it does not: the control plane stops dispatching RFC
+2136 record ops for a group with views — an nsupdate to loopback cannot target
+a view — and propagates record changes by re-rendering the zone file instead.
+``reconfig`` never reads that file, so every record created after the zone's
+first load was written to disk and never served.
+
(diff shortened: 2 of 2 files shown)


amoona6:
Closing as a duplicate of #841, which was opened earlier today from the same investigation and supersedes this.

#841 carries a third commit this PR does not: the freeze/reload/thaw here fires across **every** rendered zone on **every** record change, and a views group re-renders per record — an rndc storm during the deep test tiers. #841 diffs rendered bytes against `rendered.prev` and reloads only the zones that actually moved.

I opened this from a stale checkout of the shared QA branch and did not check for an existing upstream PR first. Apologies for the noise.

PR #843, opened 2026-08-07
fix(dns-agent): close the duplicate-`named` spawn race for real

### LINKED ISSUE #704: dns-agent: BIND9 agent starts named twice — SO_REUSEPORT hides it, rndc becomes non-deterministic (query logging silently dead)

amoona6:
Reopens and fixes #704 — the BIND9 agent still starts `named` twice.

#704 was closed as complete, but the defect reproduces on current `main`. The
mitigation that landed at the time (`drivers/_process.find_running_daemon`) says
so itself:

> the trigger is not fully established

It is now established, and this PR closes it.

## The race

`find_running_daemon` matches on `/proc/<pid>/comm`. Between `Popen(["named",
…])` returning and the child completing `execve`, that child still carries the
**forking** program's name. A concurrent caller therefore sees no daemon and
spawns a second one. That is exactly the ~118 ms window the existing docstring
describes but could not name.

Checking harder cannot fix this — the check is racing the kernel, not another
check. So the guard has to span the whole operation: `spawn_guard` takes an
exclusive `flock` held across the check, the spawn, **and** `wait_for_daemon`
(the new process becoming visible under its own name). The window then has no
interior. Both the bind9 and powerdns drivers take it; where flock is
unavailable it degrades to the previous behaviour rather than failing.

## Observed live

Inside the DNS pod on a nested 3-node appliance cluster:

```
PID   PPID  ELAPSED COMMAND
    1     0  2h02   /sbin/tini -- /usr/local/bin/spatium-dns-entrypoint
    7     1  2h02   {spatium-dns-age} /usr/bin/python3 /usr/local/bin/spatium-dns-agent
   14     7  2h02   named -f -c /var/lib/spatium-dns-agent/rendered/named.conf
   24     7  2h02   named -f -c /var/lib/spatium-dns-agent/rendered/named.conf
```

Two daemons, same parent, same config. `/run/named/named.pid` tracks only 24.
`netstat` shows 8 duplicate LISTEN rows on `:53` and 2 on `:8053` — SO_REUSEPORT
hides the collision, which is why this survives boot looking healthy.

The operational consequence is that `rndc` becomes non-deterministic. Five
successive `zonestatus` calls for one zone alternate between two authoritative
answers:

```
[0] serial: 2026080624 | nodes: 7 | last loaded: 16:06:34 | frozen: yes
[1] serial: 2026080614 | nodes: 7 | last loaded: 15:09:38 | frozen: no
[2] serial: 2026080614 | nodes: 7 | last loaded: 15:09:38 | frozen: no
```

Any control-plane operation that reloads, freezes, signs or inspects a zone
lands on whichever daemon answers, and half of them silently address the wrong
one.

| | before | after |
|---|---|---|
| `named` processes in the pod | **2** (pids 14, 24) | **1** (pid 14) |
| `rndc zonestatus` ×3 | two alternating answers | identical ×3 |

## Tests

`agent/dns`: **177 passed, 8 skipped** with this commit applied to `main`
standalone. `ruff check` introduces no new findings.

## Relationship to #838

#838 is a different defect in the same file — `swap_and_reload` sending `rndc
reconfig`, which never re-reads a changed zone file. The two were found in the
same investigation and their symptoms compound (a reload aimed at the wrong
daemon is a reload that did not happen), but they are independent: disjoint
regions of `bind9.py`, and either commit applies to `main` without the other.

### DIFF
--- a/agent/dns/spatium_dns_agent/drivers/_process.py
+++ b/agent/dns/spatium_dns_agent/drivers/_process.py
@@ -36,7 +36,12 @@ exactly the PowerDNS failure it exists to prevent.
 
 from __future__ import annotations
 
+import contextlib
+import errno
+import fcntl
 import os
+import time
+from collections.abc import Iterator
 
 
 def is_zombie(pid: str) -> bool:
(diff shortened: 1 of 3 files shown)


amoona6:
Closing as a duplicate of #841, which fixes this defect (the `spawn_guard` flock) together with the `rndc reconfig` defect, and additionally fixes an rndc storm this branch's reload introduces. Opened from a stale checkout without checking for an existing upstream PR — apologies for the noise.

Issue #641, opened 2026-07-19
DNS: operator-configurable dynamic-update (RFC 2136) ACLs on zones — TSIG + IP based, all backends


waza-ari:
## Summary

Add an operator-configurable **dynamic-update (RFC 2136) ACL** on DNS zones, so an operator can declare *"accept dynamic DNS updates to this zone from these clients"* — identified **by TSIG key** or **by source IP/CIDR** (both must be supported). Covers every supported DNS backend, degrading cleanly where a backend can't express the semantics, and resolves the drift problem (externally-injected records surviving the control-plane's full re-render).

### Motivation / gap

There is currently **no** operator-facing way to allow dynamic updates from a specific range or host (e.g. an AD domain controller or a DHCP server registering A/PTR). Confirmed today:

- The BIND9 `type master` stanza (`backend/app/drivers/dns/templates/bind9/zone.stanza.j2:20-39`) never renders `allow-update` or `update-policy`.
- The neutral `ZoneData` dataclass (`backend/app/drivers/dns/base.py:40-57`) has `allow_query`/`allow_transfer` but no update-ACL field.
- `DNSZone` (`backend/app/models/dns.py:582`), the zone schemas (`backend/app/api/v1/dns/router.py` ~712-820), and the REST surface expose no such field.
- The only inbound DDNS path is internal: the agent applies control-plane record ops over loopback via TSIG-signed `nsupdate` (`DNSServerGroup.tsig_key_*`, `backend/app/services/dns/agent_config.py:305-315`). Not operator-scopable by IP/key.
- Operator-managed `DNSTSIGKey` rows exist (`backend/app/models/dns.py:509`, API `backend/app/api/v1/dns/router.py:2518+`) and **are** rendered as `key {}` blocks — but there's **no zone field to attach them to an `allow-update` clause**. The "last mile" is missing.

> ⚠️ **Docs bug to fix as part of this:** `docs/SHIPPED.md:1546-1567` claims operators "reference keys from a zone's allow-update / allow-transfer fields as `key keyname.;`" and lists only a dropdown picker as deferred. The `allow-update` half was never implemented — only `allow_transfer` (and `allow_query`) exist on the zone model/schema/template. The prose advertises a feature that isn't there.

This is **not** the internal agent loopback path (stays as-is) and **not** the Windows AD GSS-TSIG "Secure" zone mode (already exists). It adds a third-party-writer authorization surface on top.

---

## Driver capability model (load-bearing abstraction)

[...]

Issue #61, opened 2026-05-03
Config-drift report (full record diff)


mzac:
- ⬜ **Config-drift report (full record diff)** — extends the
  existing zone-serial drift surface with a full record-level
  diff: AXFR the zone from each member, diff against the
  SpatiumDDI source of truth, surface adds / changes / deletes
  per server. Lets operators spot manual changes made directly
  on a BIND9 host.

mzac:
**Triage cross-link (2026-06-07) — decide together:** #56, #79

_Time-travel queries, config-drift report, and field-level history all read history of mutations — overlapping data model and UI._

PR #860, opened 2026-08-13
Fix DNS drift on agent-managed BIND9 — TSIG-signed AXFR + allow-transfer rendering (#734)

### LINKED ISSUE #734: DNS drift report can never succeed against agent-managed BIND9 — unsigned AXFR + inert allow_transfer
## Problem

The #61 config-drift report AXFRs each server in a group and diffs it against the DB. After the hostname-resolution fix in #61's PR, the transfer now actually reaches the server — and BIND9 answers `REFUSED`, on every zone.

Live-verified against the dev stack:

```
AXFR of example.com. from dns-bind9:53 failed: Zone transfer error: REFUSED.
```

So the Drift tab works for Windows DNS Path B and the cloud drivers (which pull over their own APIs, not AXFR), but shows nothing but "Pull failed" for the flagship agent-managed BIND9 deployment.

### 1. The control plane doesn't sign its AXFR

`backend/app/drivers/dns/_axfr.py` issues an unsigned transfer. The BIND9 agent grants `allow-transfer` **only** to the group TSIG key, and **only** on dynamic-update zones:

https://github.com/spatiumddi/spatiumddi/blob/main/agent/dns/spatium_dns_agent/drivers/bind9.py#L640-L644

```python
allow_transfer = (
    f'allow-transfer {{ key "{tsig_key_name}"; }}; '
    if (zone.get("dynamic_update_enabled") and tsig_key_name)
    else ""
)
```

The key material already exists on `DNSServerGroup.tsig_key_name` / `tsig_key_secret`, and the RFC 2136 write path already builds a keyring from it (`drivers/dns/bind9.py`, `dns.tsigkeyring.from_text`). The AXFR path just never got the same treatment — `axfr_zone_records` takes only `host` / `port` / `zone_name` and has no group context to sign with.

### 2. `allow_transfer` is settable in the UI but inert for BIND9

`DNSServerOptions.allow_transfer` (`models/dns.py:392`) and `DNSZone.allow_transfer` (`models/dns.py:748`) are both editable via the API and ride the config bundle. Neither is ever rendered: the agent's `NAMED_CONF_SKELETON` has no `allow-transfer` placeholder in the options block, and the zone stanza only emits the TSIG grant above.

Reproduced: set the dev group's `allow_transfer` to `["any"]`, wait for the bundle, and the rendered `named.conf` options block contains no `allow-transfer` line. The operator gets a 200, sees the value persisted on re-read, and nothing happens. Silent no-op.

## Suggested shape

1. Plumb the group TSIG key into `pull_zone_records` so the control plane signs the transfer (mirrors what the write path already does). Needs a driver-signature or preloaded-attribute decision, since `compute_zone_drift` doesn't eager-load `DNSServer.group`.
2. Have the agent grant `allow-transfer { key "<group-tsig>"; }` on **all** zones, not just dynamic ones — narrow, since it's key-gated rather than address-gated.
3. Render `DNSServerOptions.allow_transfer` / `DNSZone.allow_transfer` into the config, or remove the fields if they're not meant to be honoured. Either is fine; the current state — accepted and ignored — is the bad one.

Worth a look at the security posture in review: (2) opens AXFR to anyone holding the group key, which today is also the dynamic-update key. A separate transfer key may be the better shape.

## Not urgent, but it gates a shipped feature

#61's backend landed in `2026.06.11-1` and its UI ships with the PR referencing this issue. Until this is fixed, that tab is honest but empty for BIND9 users — it tells them the transfer was refused and points at `allow-transfer`, which they then can't fix, because the agent owns `named.conf`.

Found during the #61 UI work; deliberately left out of that PR since it spans the agent, the config bundle, and transfer-policy posture.

mzac:
The #61 record-level drift report and sync-with-servers both AXFR the live zone. Against **agent-managed BIND9 — the flagship deployment — every transfer was REFUSED, on every zone, 100% of the time, for two releases.**

This is a re-open: #734 was auto-closed by PR #749's comma-separated `Closes` list, which only closes the first reference. It was never fixed.

## Two defects that had to meet in the middle

**The control plane transferred unsigned.** `axfr_zone_records` had no key parameter at all, while the agent grants `allow-transfer` to the group's TSIG *key* rather than to a source address — behind an HA VIP the request can arrive from any control-plane node, and on the appliance the address isn't knowable at render time.

**The agent granted transfer on *dynamic* zones only** (a narrow #641 ingest-back grant), leaving every static zone at BIND's `none` default — and static zones are most of them. Separately, `DNSServerOptions.allow_transfer` and `DNSZone.allow_transfer` were both settable, persisted, and **never rendered at all**: the operator got a 200, saw the value on re-read, and nothing happened.

## Backend

- **`_axfr.py`** takes an optional `TsigKey`. The `dns.tsig.Key` is built explicitly — `tsigkeyring.from_text` yields `{name: bytes}` and carries no algorithm — and probed with `get_context`, because an unknown algorithm is a perfectly valid DNS *name* and nothing else rejects it until mid-transfer. An unusable key **raises rather than degrading to an unsigned transfer**, which would turn a fixable config error into the same permanent REFUSED this exists to fix.
- **New `services/dns/tsig.py`** resolves the key the agent actually granted: legacy group key first, then operator `DNSTSIGKey` rows by name, matching the bundle's ordering. Hence `op_keys` is now `order_by(name)` — unordered, two agents in one group could render different configs from the same bundle.
- **`transfer_needs_tsig()`** gates on `driver in AXFR_TSIG_DRIVERS` **and** `agent_id is not None`. The driver alone is not sufficient: a `bind9` row can be an operator's own server, authorised by address, whose `named.conf` never defined our key.
- A **keyless agent-managed group reports `unsupported` naming the missing key**, rather than a REFUSED pointing at a `named.conf` the agent owns and the operator can't edit.
- `_hint()` gained a TSIG branch matched on exception **type** — dnspython's messages carry none of the RCODE mnemonics, so a text match would have silently never fired.

## Agent — BIND9

The key grant moves to the **options block**, so it covers every zone type instead of only dynamic ones. A zone-level override re-includes the keys, since in BIND a zone clause shadows options completely — without that, setting an override would silently re-break drift for that zone. Operator ACL entries are unioned in, `none` is dropped when anything else is present, and a group with neither key nor ACL still renders closed.

## Agent — Technitium

`Deny` now falls back to `Allow` + `zoneTransferTsigKeyNames` when the group has keys.

**I nearly shipped this on an unverified assumption.** It is only safe if naming keys makes a signature *required* rather than merely permitted — and the only evidence in the repo was a claim inside an error-message string. Checked against upstream `DnsServer.cs`: `IsZoneTransferAllowed` and `IsTsigAuthenticated` are ANDed, and the second returns "no auth needed" only for an empty key set. So this **requires** a signature — narrower than an address ACL, not wider. Had it gone the other way, this would have opened AXFR of every zone on every Technitium install.

Key names are now **always sent, including empty**: upstream reads an absent parameter as "leave unchanged" and an empty one as "clear", so omitting it stranded stale names when a group's last key was deleted, and transfers from a permitted network would then fail forever demanding a signature by a key that no longer existed. Per-zone `allow_transfer` is honoured here too — otherwise the field this PR starts shipping would be a silent no-op on Technitium, the same defect one driver over.

That same source check found a live bug: `_ZONE_TRANSFER_VALUES` spelled the combined enum `AllowOnlyZoneNameServersAndUseSpecifiedNetworkACL`; upstream has no `Only` in that variant. Inert today, but it's a validator whose entire job is correctness, and Technitium silently ignores values it doesn't recognise.

## Verified live, not just unit-tested

Against the dev BIND9 on a stock `allow_transfer: ["none"]` group:

```
unsigned (the #734 bug)    FAIL — TransferError: Zone transfer error: REFUSED
signed w/ group key        OK  — 7 nodes
signed w/ WRONG secret     FAIL — PeerBadSignature
signed w/ wrong algo       FAIL — PeerBadKey
```

That last pair caught a flaw in my own error-hint code, since fixed. And the check that the `agent_id` gate matters:

```
unsigned (operator-run posture)              OK (7 nodes)
signed w/ key the server doesn't define      FAIL — PeerBadKey
```

Rendered config passes real `named-checkconf`. The drift endpoint reports `dns-bind9 → ok, in_sync=5, drift=1`, correctly catching a hand-added `ns1` record.

`make ci` green; 40 backend + 210 agent tests pass. Dev DB restored to its original state.

## Note for the next release

**Both halves of this need a new `dns-bind9` (and `dns-technitium`) agent image.** An operator who upgrades only the control plane gets signing with nothing granting it, and will reasonably conclude the fix didn't work.

## Judgement call worth flagging

A server row whose agent **hasn't registered yet** reads as not-agent-managed, so it gets an unsigned attempt rather than the fail-closed message. Deliberate — until the agent checks in there is no agent-rendered `named.conf` to have granted anything — and documented in the predicate's docstring.

Closes #734


mzac:
## QA follow-up: the headline path now has real automated coverage

Your eval found no regressions attributable to this PR, with one honest caveat — the harness never exercises the headline path, and the single check aimed at it has skipped on every tag ever walked.

I chased that down, and it's worse than "your wall doesn't cover it". **Nobody's did.**

The check is `agent/dns/tests/test_acceptance.py::test_helm_chart_primary_secondary_axfr`, a permanent `pytest.skip` whose docstring claimed:

> CI coverage lives in `.github/workflows/agent-e2e.yml`

That workflow only ever ran `dig version.bind CH TXT` — a liveness smoke, not a transfer. And `grep -rl 'axfr' .github/workflows/` matched **nothing**. There was zero automated zone-transfer coverage anywhere in the repo, and a stub confidently asserting otherwise is a large part of why #61's drift report could ship broken against every agent-managed BIND9 and stay broken across two releases.

### What's added

`agent/dns/tests/live_axfr_check.py`, run by a new **TSIG zone-transfer check** step in `build-dns-images.yml` against the bind9 image that job just built. It renders a config with the real `Bind9Driver`, starts the real pinned `named` on it, and asserts six outcomes over a real socket:

```
PASS  signed with the group key returns the zone      — 4 nodes
PASS  signed with a second granted key also works     — 4 nodes
PASS  unsigned is REFUSED                             — TransferError
PASS  wrong secret is rejected                        — PeerBadSignature
PASS  wrong algorithm is rejected                     — PeerBadKey
PASS  a key the server never granted is rejected      — PeerBadKey
```

The zone under test is deliberately **not** dynamic-update-enabled, because the pre-#734 grant existed only on dynamic zones — a static zone is the regression.

It lives in that job rather than `agent-e2e.yml` because that job already loads the image locally for Trivy, and because a unit test can only prove we emit the string we meant to emit — never that `named` reads it the way we believe.

### It is verified to fail

A check that can't fail isn't coverage, so I ran it both ways inside the real image. Reverting the fix by stubbing the key grant out:

```
Rendered named.conf transfer policy:
    allow-transfer { none; };

  FAIL  signed with the group key returns the zone   — TransferError
  FAIL  signed with a second granted key also works  — TransferError
  PASS  unsigned is REFUSED
  ...
2 expectation(s) failed
EXIT=1
```

The script also **never skips** — a missing `named` or an unimportable agent package exits non-zero, because "the check quietly didn't run" is the precise failure mode being replaced. The remaining kind-dependent skip is rewritten to state what it does and doesn't cover, and to point here.

### Also in this push

- Merged `main` (the branch was `BEHIND` after #863 landed).
- Re-ran the full live matrix on current branch code with the group in genuinely stock posture — options `allow_transfer: ["none"]`, zone override cleared, so the key grant is the *only* thing permitting a transfer. Unsigned REFUSED, signed returns 7 nodes, and the drift endpoint reports `dns-bind9 → ok, in_sync=5, drift=1`. Dev DB restored afterwards.

`make ci` green. So the confidence on this path no longer rests on a manual matrix in a PR body.

Issue #565, opened 2026-07-04
Celery worker/beat should detect DB schema behind bundled migrations (loud warn + alert), like /health/ready does for the API


mzac:
### Area

Platform / Celery / Migrations

### Motivation

An operator's local docker env logged the same 500 ~2440× in a tight loop:

```
asyncpg.exceptions.UndefinedColumnError: column platform_settings.apt_unattended_origins does not exist
```

Root cause was ordering, not data: the `api` container came up with **new model code** (the #164 unattended-upgrades columns on `PlatformSettings`) while the DB was still on the **older schema** — i.e. the app served traffic before `alembic upgrade head` ran. The DHCP agent config long-poll hammers `db.get(PlatformSettings, 1)` on a tight retry, so it stacked up thousands of identical 500s until migrations were applied. This is a generic "code deployed ahead of schema" footgun, easy to hit in dev (rebuild / `make up` before `make migrate`) and possible in prod during a botched rollout.

### The gap

- The **API** already guards against this: `/health/ready` reads the bundled Alembic head (`ScriptDirectory.get_current_head()`) and compares it to `SELECT version_num FROM alembic_version`, with distinct operator-actionable detail strings (`_check_schema_ready()` in `backend/app/api/health.py:136`; #299 / #301). In k8s this keeps the pod out of the Service until the migrate Job lands head.
- **Celery worker + beat have no equivalent.** They start on whatever schema is present and fail tasks silently in the background — the same class of failure as this incident, but with no readiness gate in front of them.

### Proposal

Reuse the existing version-vs-head comparison for Celery:

1. **Extract** the core of `_check_schema_ready()` into a shared, framework-agnostic helper (e.g. `app/core/schema_check.py: schema_at_head() -> (ok, expected, actual, detail)`), and have both `health.py` and the new callers use it. Pure DB + bundled-migration-files introspection — **no docker/k8s-specific code**, so it behaves identically in both environments (this was an explicit requirement).
2. **Worker/beat startup check** — on `worker_ready` / `beat_init` (or `@worker_process_init`), run the check once and log a loud structured warning if the DB is behind the bundled head.
3. **Periodic re-check** — a lightweight beat task (piggyback the existing 60s alert tick or a slower cadence) that re-runs the check and emits a structured warning + an `AlertEvent` on drift. Covers the "worker started before migrate finished" and "worker running a stale image" cases where a one-shot startup check would miss a later divergence.
[...]

Issue #496, opened 2026-07-02
ipam: dns_split_horizon never serialized on blocks — toggle silently flips itself off


mzac:
Found by the 2026-07-02 IPAM review sweep.

**Severity: high**

`_block_to_response` (`backend/app/api/v1/ipam/router.py:2652-2690`) hand-rolls the serialization dict and omits the `dns_split_horizon` key, so `IPBlockResponse` always emits its default `False` even when the column is `True`.

**Scenario:** operator enables split-horizon on a block → save succeeds → re-open the edit modal → toggle shows off; re-saving the form writes the column back to `False`. The block-level split-horizon inheritance feature (#25) is effectively invisible/unstable through the API.

Fix: include the field in `_block_to_response` (and audit the dict against `IPBlockResponse.model_fields` so future columns don't silently drop the same way).

Issue #299, opened 2026-05-24
Cold-boot 502 window during multi-node CNPG bootstrap + migrate


mzac:
## Symptom

On a cold-boot of a fresh appliance ISO install, there's a ~1-2 minute window where the operator can reach the web UI at the appliance IP but every API call (including the very first login attempt) returns:

```
HTTP 502 Bad Gateway
nginx/1.30.1
```

Observed on a fresh-install appliance (PR #298 ISO, dev-3afd9e4-77c8) at 192.168.0.199 a few minutes after first boot:

- Host nginx (the frontend pod with `hostNetwork: true` + hostPorts 80/443) is up and serving requests.
- `api` pod is Running and answering `/health/live` 200 OK.
- `migrate` is still running — saw two distinct job pods complete with `~3 min` gap between them (`migrate-r1` at 4m02s, `migrate-r2` at 108s).
- CNPG bootstrap operator pod showed `RESTARTS 2 (93s ago)` before the primary stabilized.

Once `migrate-r2` completed and the CNPG primary settled, `admin/admin` login worked normally.

## Why it happens

The frontend nginx happily proxies `/api/v1/auth/login` through to the api pod the moment its ClusterIP is reachable. The api FastAPI process starts as soon as the container starts (it doesn't gate on DB readiness), so `/health/live` returns 200. But the api's actual route handlers can't answer because:

1. CNPG primary takes a couple of restart cycles to become ready on a multi-node bootstrap (#272 work).
2. `migrate` retries until the schema lands — so even after CNPG is up, there's a brief window before the schema actually exists.
3. The api responds to the login query by timing out against the not-yet-migrated DB, and nginx surfaces that as a bare 502.

## What an operator sees

A bare nginx 502 page with no context — looks like a real failure, not a "still booting" condition. Easy to assume the install is broken and reach for the docs / open an issue / reinstall.

## Suggested fixes

Pick the smallest one that closes the gap; can pick more than one:

1. **Tighten `/health/ready` to do a real schema query** — currently liveness and readiness probes both pass on FastAPI's process being up. If readiness gated on `SELECT 1 FROM alembic_version` (or similar lightweight schema-aware query), kubernetes wouldn't add the api pod to the Service endpoint set until migrate has completed, and nginx would return its own "no upstream" message during the gap instead of routing into a doomed-to-timeout backend.
[...]

Issue #123, opened 2026-05-07
Diagnostics — capture uncaught exceptions + admin error viewer + "Submit bug" GitHub prefill


mzac:
SpatiumDDI today swallows uncaught Python errors into the API + worker stdout streams. Operators discover them by tailing `docker compose logs -f api worker`, which means: nobody sees the error unless they happen to be tailing at the right moment, and there's no path from "I see a traceback" to "I've filed a bug for it" beyond manual copy-paste. This issue adds a diagnostic surface that captures every uncaught exception, persists it for review, and offers a one-click GitHub-issue prefilled with the traceback.

## Capture

A new `internal_error` table:

| column | type | notes |
|---|---|---|
| `id` | UUID | PK |
| `timestamp` | timestamptz | server-side `now()` |
| `service` | text | `api` / `worker` / `beat` |
| `kind` | text | `unhandled_exception` (today's only kind; reserved for future expansion) |
| `request_id` | text | from structlog context where available |
| `route_or_task` | text | request path + method, or Celery task name |
| `exception_class` | text | e.g. `sqlalchemy.exc.IntegrityError` |
| `message` | text | one-line exception message |
| `traceback` | text | full traceback, max 16 KB, truncated with marker if longer |
| `context_json` | jsonb | sanitised request context: user_id, params (no secrets), headers (no Auth), task args |
| `acknowledged_by` | UUID | nullable, FK to user |
| `acknowledged_at` | timestamptz | nullable |
| `suppressed_until` | timestamptz | nullable; while ≥ now(), matching errors increment `occurrence_count` instead of inserting a new row |
| `fingerprint` | text | sha256 of `(exception_class, top-2-frames-from-traceback)` for grouping |
| `occurrence_count` | int | default 1; bumps on duplicate fingerprints within the suppression window |
| `last_seen_at` | timestamptz | last occurrence |

Capture sites:

- **API**: a FastAPI exception handler registered last (after auth + permission handlers). Handles non-HTTPException raises by inserting an `internal_error` row, then re-raising so the existing 500 handler still returns to the client.
- **Worker**: `celery.signals.task_failure.connect(...)` — captures uncaught task exceptions. `celery.signals.task_revoked` and `task_unknown` not in scope (those are operational, not bugs).
- **Beat**: same `task_failure` signal covers it (beat dispatches through the same broker).

[...]

PR #861, opened 2026-08-14
fix(api): probe budgets + thirteen 5xx-class defects found by conformance fuzzing

amoona6:
## Summary

Schemathesis conformance fuzzing (auth'd, all operations) plus deep-tier API
tests against real appliance deployments turned up one operational defect and
thirteen 5xx-class API defects. Every item below was observed live on a k3s
appliance before it was fixed, and the full set was re-proven green on an
appliance built from this branch (final fuzz run: 1017 pass / 0 fail, and a
full evaluation walk with the UI suite 145/145).

### 1. The api pod's probes have no `timeoutSeconds` (the keystone)

`charts/spatiumddi/templates/api.yaml` and `k8s/base/api.yaml` set
initialDelay/period/failureThreshold on both probes but not `timeoutSeconds`
— the only workload in the chart without it — so both probes inherit
Kubernetes' 1-second default. The api is a single-worker uvicorn; bcrypt
verification, inline reportlab PDF rendering, and heavy estate queries all
run on the event loop and push the probe tail past 1s under ordinary load.

Observed live (idle appliance, healthy api pod, 0 restarts):

```
Warning  Unhealthy  pod/...-api-...  Liveness probe failed:
  Get "http://<pod>:8000/health/live": context deadline exceeded
Warning  Unhealthy  pod/...-api-...  Readiness probe failed:
  Get "http://<pod>:8000/health/ready": context deadline exceeded
```

Three straight tail probes pull the pod from the Service (readiness) or
restart a healthy container (liveness). Operators see recurring ~1-minute
windows where nginx answers 502 for every route — and serves the SPA's
index.html for `/openapi.json`. Under sustained load this recurred every
5–15 minutes; after the fix an appliance rode out a full fuzz run with zero
probe events.

Fix: explicit `timeoutSeconds: 5` on both probes (liveness failureThreshold
3→4), and the reportlab render pass moves to `asyncio.to_thread` in both PDF
services so an export cannot stall the probe path.

Considered and not shipped: `--workers 2` in the api Dockerfile — the
DNS/DHCP agent long-poll wake is an in-process `asyncio.Event`, so a second
worker would silently degrade config-push latency. Worth its own change with
a cross-process wake.

Related follow-up (not in this PR): under `podAntiAffinity: hard` the chart
deliberately uses `maxSurge: 0, maxUnavailable: 1` (the surge pod deadlocks
against the outgoing ReplicaSet — the wedge the helper's comment documents).
On a 2-replica api that guarantees a no-ready-endpoint window on every
template reconcile; `matchLabelKeys: [pod-template-hash]` on the
anti-affinity term (stable since k8s 1.29) would make surge-first safe, but
deserves its own PR with a 1→3 promote test.

### 2. `verify_password` raised instead of answering False

`bcrypt.checkpw` raises `ValueError` for >72-byte or NUL-containing
candidates; request schemas allow 256 chars. Every password-gated endpoint
500'd on such input — `DELETE /appliance/appliances/{id}` and
`POST .../permanent-delete` failed every conformance run on three
consecutive builds; login/reauth/factory-reset carry the same latent path.
bcrypt reads only the first 72 bytes of a valid input, so the truncation
cannot accept a password the untruncated compare would reject.

### 3. The delete denial path crashed writing its own audit row

`AuditLog(error_message="bad_password")` — `error_message` is not a column
(`error_detail` is) — so the wrong-password DENIAL path raised TypeError and
answered 500 (16 of 77 tracebacks in one fuzz run's pod log; reproduced by
hand with an ordinary wrong password). This sat behind item 2 on the same
path — both were required. `resource_display` (NOT NULL) rides along.

### 4. `GET /admin/postgres/schema-health` has answered 500 since #565

The handler's lazy import kept the pre-#565 path
(`app.api.health._expected_alembic_head`); #565 moved the reader to
`app.core.schema_check.expected_alembic_head`. ImportError fired before any
of the handler's own `status="error"` fallbacks could run — a 500 on a
perfectly healthy appliance, every call. Regression test pins the endpoint
to never-500.

### 5. export.pdf operations: media type sent but not declared

Both PDF exports send `Content-Type: application/pdf` correctly, but a bare
`-> Response` documents the 200 as application/json only, so schema-aware
clients flag every response (`UndefinedContentType`). Declared now.

### 6. `do_not_probe` echoed and re-read as `false`

`_block_to_response` omitted `do_not_probe`/`do_not_probe_reason` — the same
defect class as the #496 `dns_split_horizon` omission it sits next to. The
column persisted correctly; every read lied, and a UI round-trip wrote the
False back. A block the operator excluded from probing read as
probe-eligible everywhere.

### 7. Agent register: unbounded strings into bounded columns

`AgentRegisterRequest*` (dns/dhcp/looking-glass) declared bare `str` fields
landing in `String(255)/(128)/(50)` columns — over-length input surfaced as
asyncpg `StringDataRightTruncation` → 500. The models now mirror their
columns and answer 422.

### 8. The agent PSK gates crashed on a non-ASCII key

`hmac.compare_digest` raises TypeError for a str containing non-ASCII
characters, so all three bootstrap gates (`X-DNS/DHCP/LG-Agent-Key`)
answered 500 to such a key (fuzz case: `'\x80'`) — the AUTH layer itself was
the crash. Both sides are now surrogateescape-encoded bytes: same
constant-time comparison, same accept set, and no input can throw.
Regression test drives all three gates with the failing byte.

### 9. audit-forward-target `{target_id}` typed `str`

A non-UUID path value (`'0'`, generated constantly) reached asyncpg's bind
and died as a 500 on update/delete/test. Typed `uuid.UUID`, FastAPI answers
the 422 it is.

### 10. new-devices/block accepted any string into a MACADDR column

`BlockBody.mac_address: str` with no validation — `""` died at bind as
`InvalidTextRepresentationError` → 500. The repo already has the exact guard
for this (`app/api/v1/dhcp/_mac.py`, whose docstring records the identical
defect in the reservation form); the same shared `canonicalize_mac` now
answers 422.

### 11. Data-shaped driver errors are 422, constraint conflicts are 409

Check-then-insert handlers (asns, agent register's group auto-create) race
between the SELECT and the flush; the loser's unique violation surfaced as
500 where the pre-check's own answer was 409 — a global `IntegrityError`
handler answers it. Values Postgres refuses as DATA (NUL escapes → 22P05,
malformed macaddr/uuid literals → 22P02, over-length → 22001) answer 422 via
a `DBAPIError` handler keyed on SQLSTATE class 22 / asyncpg's DataError
lineage — chosen because the asyncpg dialect wraps these as GENERIC
`DBAPIError`, and anything non-data (e.g. a ProgrammingError in our own SQL)
re-raises and stays the 500 it is.

### 12. CSR `country` validated in characters, consumed in bytes

`CSRGenerate.country` used `min_length=2, max_length=2` — pydantic counts
characters, cryptography's `CountryName` counts encoded bytes, so a 2-char
multibyte string passed validation and blew up the x509.Name build as a 500
("length must be >= 2 and <= 2, but it was 6"). The field is now the two
ASCII letters the x509 spec means.

### 13. Import connection fields accept only what the wire can carry

The PowerDNS/Technitium/NetBox connection fields (`api_url`, `api_key`,
`base_url`, `token`) accepted arbitrary unicode, but a URL and an HTTP
header are ASCII on the wire — the outbound httpx client crashed on the
request build (UnicodeEncodeError; httpx "Invalid HTTP header value") as a
500 before any pull started. Printable-ASCII patterns make them the 422 they
are.

### 14. The import previews' documented 502 is now declared

The seven DNS/NetBox import test-connection/preview operations answer 502 by
design when the operator's remote source is unreachable (gateway semantics —
the operator's PowerDNS/NetBox is the upstream). The status is now declared
on those operations so schema-aware clients can tell the documented answer
from an infrastructure 502.

## Area

- [x] DNS
- [x] DHCP
- [x] Auth / Providers
- [x] Audit / Observability
- [x] API
- [x] Deployment (Compose / K8s / Appliance)
- [ ] IPAM (one response-serializer fix, item 6)

## Screenshots / API examples

Before → after on the same appliance build line:

```sh
# 4. schema-health (was: 500 ImportError on every call)
$ curl -sk https://<appliance>/api/v1/admin/postgres/schema-health -H "Authorization: Bearer $TOK"
{"status":"ok","expected_head":"...","db_revision":"...","detail":"..."}   # was: {"detail":"Internal Server Error"}

# 6. do_not_probe (was: echoed and re-read false)
$ curl -sk -X POST https://<appliance>/api/v1/ipam/blocks -d '{"space_id":"...","network":"10.9.0.0/16","name":"x","do_not_probe":true}' ...
{... "do_not_probe": true ...}   # was: "do_not_probe": false

# 2+3. appliance delete with a wrong password (was: 500)
$ curl -sk -X DELETE https://<appliance>/api/v1/appliance/appliances/<id> -d '{"password":"wrong"}' ...
403 {"detail":"Current password incorrect."}   # was: 500, and no denial audit row
```

## Test plan

- [x] `make ci` passes (backend ruff+black+mypy, frontend lint, frontend build)
- [x] New/changed behavior is covered by a test — new:
      `tests/test_admin_postgres_schema_health.py` (endpoint never-500,
      `verify_password` totality), `tests/test_agent_psk_gate.py` (all three
      gates × the live failing byte), `tests/test_fuzz_input_validation.py`
      (CSR country byte-counting, import-credential ASCII); targeted runs of
      `tests/test_audit_forward.py`, `tests/test_new_device_watch.py`,
      `tests/test_health_schema_check.py`, `tests/test_soft_delete.py` green
- [x] Manually verified via curl and via schemathesis conformance runs
      against real k3s appliance deployments: baseline build 987 pass /
      14-per-run fail on exactly the defects above; this branch 1017 pass /
      0 fail on the same fuzz, api pod 0 restarts and zero probe-timeout
      events through the run; a full evaluation walk (install → upgrade
      cycle → conformance+deep API → UI 145/145 → protocol/cluster suites)
      green on every suite this PR touches

## Migration / deployment notes

- [ ] Includes an Alembic migration (no DB model changes)
- [x] Updated `k8s/base/` manifests (probe `timeoutSeconds`/threshold only —
      no service or env changes, `k8s/README.md` unaffected)
- [ ] Breaking change — none: every 5xx→4xx conversion is on inputs that
      previously crashed; the newly declared 502s were already answered on
      the wire

## Related issues

Refs #496 (same serializer-omission class, adjacent field), #565 (the
schema-health import drift), #299 (readiness gate interaction with the
probe budgets).


backend/app/core/security.py
```diff
@@ -22,7 +22,18 @@ def hash_password(password: str) -> str:
...
+    # Total by contract: a candidate password can only ever be WRONG, never an
+    # exception. bcrypt.checkpw raises ValueError on >72-byte input and on NUL
+    # bytes (and on a malformed stored hash) — request schemas allow up to 256
+    # chars, so any caller that fed this a user-supplied string 500'd on those
+    # inputs instead of answering 401/403 (live: DELETE
+    # /appliance/appliances/{id} answered 500 on every conformance run).
+    # bcrypt itself only reads the first 72 bytes of a valid input, so the
+    # truncation cannot accept a password the untruncated compare would reject.
+    try:
+        return bcrypt.checkpw(plain.encode()[:72], hashed.encode())
```
mzac: `verify_password` is now total, but its sibling two functions up isn't:

```python
def hash_password(password: str) -> str:
    return bcrypt.hashpw(password.encode(), bcrypt.gensalt()).decode()
```

`bcrypt.hashpw` raises on >72 bytes just as `checkpw` does — confirmed in a container:

```
ValueError: password cannot be longer than 72 bytes, truncate manually if necessary
```

And nothing bounds the input. The only validators are `min_length: 8` (`users/router.py:111` and `:148`) and `password_policy.validate`, which has `min_length` but **no maximum** (`services/password_policy.py:87`). So a 100-char password passes policy and then raises inside `_enforce_policy`.

That leaves `POST /api/v1/users`, `POST .../reset-password` and the change-password path still 500-ing on exactly the input class item 2 says it closed. There's also a duplicate `bcrypt.hashpw` at `users/router.py:186` that needs the same treatment.

Truncating to `[:72]` here mirrors what `verify_password` now does, and your own comment already argues why that's safe.

backend/app/main.py
```diff
@@ -867,6 +867,65 @@ async def _transient_db_connection(request: Request, exc: Exception) -> Response
             headers={"Retry-After": "1"},
         )
 
+    from sqlalchemy.exc import DBAPIError as SADBAPIError  # noqa: PLC0415
+    from sqlalchemy.exc import IntegrityError as SAIntegrityError  # noqa: PLC0415
+
+    @app.exception_handler(SAIntegrityError)
```
mzac, line 886: This catches **every** `IntegrityError`, not just unique violations — NOT NULL (23502), FK (23503) and CHECK (23514) come through here too. Those aren't the data conflicting; they're our code being wrong, and they now get blamed on the client.

The sharpest version of the problem is self-referential: **item 3 of this very PR was a NOT NULL violation** (`error_detail` / `resource_display`). Reintroduce it tomorrow and this handler answers 409 — so a Schemathesis run that flags 5xx passes clean. The handler makes the fuzz suite blind to the bug class the PR just fixed.

Second-order effect, which I checked: Starlette resolves handlers by exact type then MRO, so `IntegrityError` matches here and never reaches the `@app.exception_handler(Exception)` at line 934 that calls `record_unhandled_exception_async`. These also disappear from the #123 diagnostics capture.

Suggest narrowing to the unique-violation SQLSTATE, in the same spirit as the `_unstorable_value` handler below discriminating on SQLSTATE class 22:

```python
if getattr(getattr(exc, "orig", None), "sqlstate", None) != "23505":
    raise exc
```

That keeps the documented fix (the SELECT-then-INSERT races in asns and agent register) and lets genuine server faults stay 500.

mzac (commented):
Reviewed the full diff against `main`, reading each hunk plus the surrounding context (models, callers, exception-handler ordering, probe manifests). Every finding below was verified against this branch's head rather than inferred — including running `bcrypt.hashpw` in a container to confirm the raise.

**The core of the PR holds up.** I specifically checked and cleared: the `schema_check` import path; `error_detail` / `resource_display` matching the NOT NULL columns; the `Field` imports existing in all three agent routers; the shared `_ASCII` `FieldInfo` reuse being safe under pydantic 2.13 (rust-regex also rejects the trailing-newline `$` edge case); `canonicalize_mac` raising `ValueError` so it yields a real 422; `IPBlockResponse` already carrying both `do_not_probe` fields (both NOT NULL, so no serializer break); all 7 endpoints that raise 502 now being declared; and MRO-based handler lookup meaning the new `DBAPIError` handler cannot steal the transient-connection path.

Seven things I'd want addressed, left inline. Two are worth calling out here:

**`hash_password` was left behind (inline on `security.py`).** `verify_password` became total, but the password-*set* paths didn't, and nothing caps length anywhere — the validators enforce `min_length: 8` and `password_policy` has `min_length` but no maximum. Confirmed empirically that `bcrypt.hashpw` raises above 72 bytes. So create-user / reset-password / change-password still 500 on the exact input class item 2 says it closed.

**The blanket `IntegrityError` → 409 interacts badly with this PR's own methodology (inline on `main.py`).** It catches NOT NULL and FK violations, not just unique conflicts. Item 3 of this PR *was* a NOT NULL violation from our own code; reintroduced tomorrow it would now answer 409 and a Schemathesis run that flags 5xx would sail straight past it. Second-order effect I confirmed: Starlette dispatches exact-type-then-MRO, so `IntegrityError` never reaches the `@app.exception_handler(Exception)` at `main.py:934` feeding `record_unhandled_exception_async` — those bugs also vanish from the #123 diagnostics surface.

**One finding has no inline anchor** because the file isn't in the diff:

`backend/app/services/ipam_io/pdf.py:597` — the keystone fix moved two reportlab renders to `asyncio.to_thread`, but the IPAM `export.pdf` renderer, the largest of the three, still calls `doc.build()` inline on the event loop. A large tree export therefore still stalls the probe path this PR exists to protect, past even the new `timeoutSeconds: 5`. Its route (`ipam/io_router.py:224`) also still declares only `application/json`, so it's missed by item 5 as well.

Nothing here is a blocker on the approach — the probe budget and the 5xx work are clearly right, and the live-verification standard in the description is excellent. These are gaps in coverage rather than errors in direction.

backend/app/services/conformity/pdf.py
```diff
@@ -166,7 +167,10 @@ async def generate_conformity_pdf(
                 body,
             )
         )
-        doc.build(story)
+        # reportlab's render pass is synchronous CPU work; on the single-
+        # worker uvicorn loop it stalls every request AND the kubelet's
+        # 1s-budget probes, so it runs on a worker thread.
+        await asyncio.to_thread(doc.build, story)
```
mzac, line 175: This is right, but the docstring above it (line 111) now says the opposite:

> The PDF is fully synchronous + in-memory; for large estates the callsite should run this in `asyncio.to_thread` since reportlab itself isn't async-aware.

`generate_conformity_pdf` is `async def` and does the `to_thread` internally, here and at line 309. A developer adding a second caller and following that instruction writes `asyncio.to_thread(generate_conformity_pdf, db)` — which hands a worker thread an async function, gets back an un-awaited coroutine object instead of PDF bytes, and fails silently apart from a `coroutine was never awaited` RuntimeWarning.

Same issue in `services/audit_report.py:7`, whose module docstring asserts "synchronous render after async DB queries; small enough to render inline, no `asyncio.to_thread` needed at our scale" while the module now imports `asyncio` and uses it.

backend/app/api/v1/new_devices/router.py
```diff
@@ -89,6 +91,16 @@ class BlockBody(BaseModel):
...
     block_upstream: bool = False
 
+    # Same guard (and same shared helper) as the DHCP mac-block/static
+    # models: the value lands in a MACADDR column, and an unparseable MAC
+    # used to reach Postgres and come back as a bind-time 500 (live:
+    # mac_address="" → InvalidTextRepresentationError). Rejecting here
+    # makes it the 422 it is.
+    @field_validator("mac_address")
+    @classmethod
+    def _norm_mac(cls, v: str) -> str:
```
mzac, line 101: Good — but `AllowlistCreate` further down this same file (line ~125) still has an unguarded `mac_address`:

```python
class AllowlistCreate(BaseModel):
    mac_address: str | None = None
```

No `field_validator`, so `{"mac_address": "zz"}` reaches `CAST(:mac AS macaddr)` exactly as `BlockBody` used to. It no longer 500s thanks to the new `_unstorable_value` handler, but it produces a generic field-less 422 instead of the precise per-field one this guard gives every sibling schema — so the fix is inconsistent within one file.

The `None` default means the validator needs a `None` passthrough (the `_one_key` model validator already allows either key).

backend/app/api/v1/dhcp/agents.py
```diff
@@ -64,12 +64,16 @@
...
+    # .driver 50, .agent_fingerprint 128, DHCPServerGroup.name 255) — an
+    # over-length value used to reach asyncpg and surface as 500 instead
+    # of the 422 it is.
+    hostname: str = Field(max_length=255)
+    driver: str = Field(default="kea", max_length=50)
     roles: list[str] = ["standalone"]
-    version: str | None = None
-    group_name: str | None = None
-    fingerprint: str
+    version: str | None = Field(default=None, max_length=100)
```
mzac: `max_length=100` here is wider than the column it lands in — `DHCPServer.agent_version` is `String(64)` (`models/dhcp.py:202`).

An agent reporting an 80-character version passes field validation, then trips the column on INSERT: asyncpg raises `StringDataRightTruncation`, which the new `_unstorable_value` handler turns into a generic 422. Better than a 500, but it's a bound that simply doesn't match storage — `max_length=64` makes it a precise field-level 422 instead.

Same mismatch on the looking-glass register schema, and the heartbeat models leave `agent_version` unbounded entirely, so the same value can arrive by that route with no bound at all.

charts/spatiumddi/templates/api.yaml
```diff
@@ -96,13 +96,23 @@ spec:
             {{- end }}
           # Liveness — "process responds to HTTP". Deliberately shallow:
           # a transient Postgres blip should not trigger a pod restart.
+          #
+          # timeoutSeconds is EXPLICIT because the k8s default is 1s and the
+          # api is a single-worker uvicorn: any synchronous work on the event
+          # loop (bcrypt, an inline render, a slow estate query) pushes the
+          # probe's p95 past 1s under load, and three straight tail probes had
+          # the kubelet restarting a HEALTHY pod every few minutes — operators
+          # saw it as recurring ~1 min windows of bare nginx 502s. Every other
```
mzac, line 105: > Every other workload in this chart already sets timeoutSeconds; the api was the one omission.

The frontend Deployment doesn't — `charts/spatiumddi/templates/frontend.yaml:141-148` sets `periodSeconds` on both probes but no `timeoutSeconds`, and `k8s/base/frontend.yaml:49-60` matches. Both still inherit the 1s default.

nginx serving static files is far less likely to blow a 1s budget than a single-worker uvicorn, so I don't think it's urgent — but the claim as written will stop the next reader from checking, which is how the api's own omission survived this long. Either fix the frontend in passing or soften the sentence to name it as a known remaining case.

backend/app/core/security.py
mzac: `verify_password` is now total, but its sibling two functions up isn't:

```python
def hash_password(password: str) -> str:
    return bcrypt.hashpw(password.encode(), bcrypt.gensalt()).decode()
```

`bcrypt.hashpw` raises on >72 bytes just as `checkpw` does — confirmed in a container:

```
ValueError: password cannot be longer than 72 bytes, truncate manually if necessary
```

And nothing bounds the input. The only validators are `min_length: 8` (`users/router.py:111` and `:148`) and `password_policy.validate`, which has `min_length` but **no maximum** (`services/password_policy.py:87`). So a 100-char password passes policy and then raises inside `_enforce_policy`.

That leaves `POST /api/v1/users`, `POST .../reset-password` and the change-password path still 500-ing on exactly the input class item 2 says it closed. There's also a duplicate `bcrypt.hashpw` at `users/router.py:186` that needs the same treatment.

Truncating to `[:72]` here mirrors what `verify_password` now does, and your own comment already argues why that's safe.

backend/app/main.py
mzac, line 886: This catches **every** `IntegrityError`, not just unique violations — NOT NULL (23502), FK (23503) and CHECK (23514) come through here too. Those aren't the data conflicting; they're our code being wrong, and they now get blamed on the client.

The sharpest version of the problem is self-referential: **item 3 of this very PR was a NOT NULL violation** (`error_detail` / `resource_display`). Reintroduce it tomorrow and this handler answers 409 — so a Schemathesis run that flags 5xx passes clean. The handler makes the fuzz suite blind to the bug class the PR just fixed.

Second-order effect, which I checked: Starlette resolves handlers by exact type then MRO, so `IntegrityError` matches here and never reaches the `@app.exception_handler(Exception)` at line 934 that calls `record_unhandled_exception_async`. These also disappear from the #123 diagnostics capture.

Suggest narrowing to the unique-violation SQLSTATE, in the same spirit as the `_unstorable_value` handler below discriminating on SQLSTATE class 22:

```python
if getattr(getattr(exc, "orig", None), "sqlstate", None) != "23505":
    raise exc
```

That keeps the documented fix (the SELECT-then-INSERT races in asns and agent register) and lets genuine server faults stay 500.

backend/app/services/conformity/pdf.py
mzac, line 175: This is right, but the docstring above it (line 111) now says the opposite:

> The PDF is fully synchronous + in-memory; for large estates the callsite should run this in `asyncio.to_thread` since reportlab itself isn't async-aware.

`generate_conformity_pdf` is `async def` and does the `to_thread` internally, here and at line 309. A developer adding a second caller and following that instruction writes `asyncio.to_thread(generate_conformity_pdf, db)` — which hands a worker thread an async function, gets back an un-awaited coroutine object instead of PDF bytes, and fails silently apart from a `coroutine was never awaited` RuntimeWarning.

Same issue in `services/audit_report.py:7`, whose module docstring asserts "synchronous render after async DB queries; small enough to render inline, no `asyncio.to_thread` needed at our scale" while the module now imports `asyncio` and uses it.

backend/app/api/v1/new_devices/router.py
mzac, line 101: Good — but `AllowlistCreate` further down this same file (line ~125) still has an unguarded `mac_address`:

```python
class AllowlistCreate(BaseModel):
    mac_address: str | None = None
```

No `field_validator`, so `{"mac_address": "zz"}` reaches `CAST(:mac AS macaddr)` exactly as `BlockBody` used to. It no longer 500s thanks to the new `_unstorable_value` handler, but it produces a generic field-less 422 instead of the precise per-field one this guard gives every sibling schema — so the fix is inconsistent within one file.

The `None` default means the validator needs a `None` passthrough (the `_one_key` model validator already allows either key).

backend/app/api/v1/dhcp/agents.py
mzac: `max_length=100` here is wider than the column it lands in — `DHCPServer.agent_version` is `String(64)` (`models/dhcp.py:202`).

An agent reporting an 80-character version passes field validation, then trips the column on INSERT: asyncpg raises `StringDataRightTruncation`, which the new `_unstorable_value` handler turns into a generic 422. Better than a 500, but it's a bound that simply doesn't match storage — `max_length=64` makes it a precise field-level 422 instead.

Same mismatch on the looking-glass register schema, and the heartbeat models leave `agent_version` unbounded entirely, so the same value can arrive by that route with no bound at all.

charts/spatiumddi/templates/api.yaml
mzac, line 105: > Every other workload in this chart already sets timeoutSeconds; the api was the one omission.

The frontend Deployment doesn't — `charts/spatiumddi/templates/frontend.yaml:141-148` sets `periodSeconds` on both probes but no `timeoutSeconds`, and `k8s/base/frontend.yaml:49-60` matches. Both still inherit the 1s default.

nginx serving static files is far less likely to blow a 1s budget than a single-worker uvicorn, so I don't think it's urgent — but the claim as written will stop the next reader from checking, which is how the api's own omission survived this long. Either fix the frontend in passing or soften the sentence to name it as a known remaining case.

mzac (commented):
Reviewed the full diff against `main`, reading each hunk plus the surrounding context (models, callers, exception-handler ordering, probe manifests). Every finding below was verified against this branch's head rather than inferred — including running `bcrypt.hashpw` in a container to confirm the raise.

**The core of the PR holds up.** I specifically checked and cleared: the `schema_check` import path; `error_detail` / `resource_display` matching the NOT NULL columns; the `Field` imports existing in all three agent routers; the shared `_ASCII` `FieldInfo` reuse being safe under pydantic 2.13 (rust-regex also rejects the trailing-newline `$` edge case); `canonicalize_mac` raising `ValueError` so it yields a real 422; `IPBlockResponse` already carrying both `do_not_probe` fields (both NOT NULL, so no serializer break); all 7 endpoints that raise 502 now being declared; and MRO-based handler lookup meaning the new `DBAPIError` handler cannot steal the transient-connection path.

Seven things I'd want addressed, left inline. Two are worth calling out here:

**`hash_password` was left behind (inline on `security.py`).** `verify_password` became total, but the password-*set* paths didn't, and nothing caps length anywhere — the validators enforce `min_length: 8` and `password_policy` has `min_length` but no maximum. Confirmed empirically that `bcrypt.hashpw` raises above 72 bytes. So create-user / reset-password / change-password still 500 on the exact input class item 2 says it closed.

**The blanket `IntegrityError` → 409 interacts badly with this PR's own methodology (inline on `main.py`).** It catches NOT NULL and FK violations, not just unique conflicts. Item 3 of this PR *was* a NOT NULL violation from our own code; reintroduced tomorrow it would now answer 409 and a Schemathesis run that flags 5xx would sail straight past it. Second-order effect I confirmed: Starlette dispatches exact-type-then-MRO, so `IntegrityError` never reaches the `@app.exception_handler(Exception)` at `main.py:934` feeding `record_unhandled_exception_async` — those bugs also vanish from the #123 diagnostics surface.

**One finding has no inline anchor** because the file isn't in the diff:

`backend/app/services/ipam_io/pdf.py:597` — the keystone fix moved two reportlab renders to `asyncio.to_thread`, but the IPAM `export.pdf` renderer, the largest of the three, still calls `doc.build()` inline on the event loop. A large tree export therefore still stalls the probe path this PR exists to protect, past even the new `timeoutSeconds: 5`. Its route (`ipam/io_router.py:224`) also still declares only `application/json`, so it's missed by item 5 as well.

Nothing here is a blocker on the approach — the probe budget and the 5xx work are clearly right, and the live-verification standard in the description is excellent. These are gaps in coverage rather than errors in direction.

amoona6:
All six addressed in `49bae52`. Thanks — two of these changed the shape of the fix rather than polishing it.

**1. `hash_password` (and it was worse than reported).** `_enforce_policy` in `users/router.py` called `bcrypt.hashpw` **directly**, so fixing only the helper would have left create-user and reset-password broken anyway. Every hashing path now routes through `hash_password`, which truncates identically to `verify_password` — the pair has to agree on the boundary or a >72-byte password would hash fine and then never verify. Added a real maximum to `password_policy.validate` so the operator gets an actionable message instead of a 500 from inside the hash, stated in **bytes** because that is what bcrypt measures: 40 two-byte characters are over the line while 71 ASCII ones are not, and a character count gets that backwards.

**2. The `IntegrityError` handler — you're right, and the self-referential argument is what settles it.** Now 409 only for SQLSTATE `23505`; NOT NULL / FK / CHECK re-raise to the 500 they are. A handler that turns our own bug class into a 4xx makes the conformance suite blind to exactly the regression this PR fixed, which is a worse failure mode than the 500 it was smoothing over.

**3. PDF docstrings** — corrected in both modules. The wrong instruction was the dangerous part: a second caller following it would hand a worker thread an async function and get an un-awaited coroutine instead of bytes.

**4. `AllowlistCreate`** — guarded with the same `canonicalize_mac`, with a `None` passthrough since either key is valid.

**5. `agent_version`** — 100 → 64 to match `String(64)`, across all three agent families, **and** on the heartbeat models, which were unbounded entirely; that was the side door the same value arrived through.

**6. Frontend probes** — fixed rather than softened. "nginx is too simple to blow a 1s budget" is the same reasoning that let the api's own omission survive, so both manifests now set `timeoutSeconds` explicitly.

`make ci` green; targeted backend tests 31 passed, including new regressions for the hashing totality and the byte-based policy bound.

mzac:
Re-reviewed `49bae52` hunk by hunk against the branch head. **All six are correctly fixed**, and two of them are better than what I asked for:

- **Hashing totality** — finding `_enforce_policy` calling `bcrypt.hashpw` directly was the part I missed; patching only the helper would have left create-user and reset-password broken exactly as before. The shared `BCRYPT_MAX_BYTES` and the byte-stated policy maximum are the right shape, and stating it in bytes is the detail most implementations get wrong. I traced the ordering on both set paths — `validate` → `is_in_history` → `hash_password` at `auth/router.py:826-846` and `users/router.py:180-186` — so the cap genuinely gates every path with no side door.
- **`agent_version`** — catching that the heartbeat models were unbounded *entirely* closed the route the value actually arrived by. That was the more important half.

Two things I checked this pass and cleared, so nobody re-opens them: `core/acme_auth.py` uses bcrypt directly, but `hash_acme_password` only ever receives a server-generated 40-char secret and `verify_acme_password` already catches `ValueError`/`TypeError`; and `password_policy.is_in_history` calls `checkpw` untruncated but sits inside a total `except`, behind the new byte cap on both paths.

### The seventh finding — fixed in `0344b0d4`

The one with no inline anchor, because the file wasn't in the diff. It didn't make it into `49bae52`, so I've pushed it rather than send you round again:

`services/ipam_io/pdf.py` still called `doc.build()` inline on the event loop — the largest of the three renderers, a two-pass numbered build over every row in the subtree, so the one most able to blow the probe budget the PR exists to protect, `timeoutSeconds: 5` included. Now `asyncio.to_thread`, matching the conformity and audit renderers. `ipam/io_router.py:224` was likewise still bare, so item 5 missed it; it now carries the same `responses={200: {"content": {"application/pdf": {}}}}` block — verified in the generated schema, `/api/v1/ipam/export.pdf` now declares `application/pdf` alongside audit's.

Both docstrings went in the same edit: the module still claimed the render was "small enough at our scale to build inline", and the function lacked the do-not-double-wrap warning its two siblings now carry.

`tests/test_ipam_pdf_export.py` + `tests/test_admin_postgres_schema_health.py` 34 passed; ruff, black and mypy clean on both files.

All twelve threads resolved (my review submitted twice by accident, hence the duplicates — sorry for the noise). Nothing outstanding from my side.

Issue #520, opened 2026-07-02
ipam: cross-subnet bulk operations from filter/search results


mzac:
From the 2026-07-02 IPAM review sweep.

Selection state (`selectedIpIds`) lives inside `SubnetDetail` — bulk edit/delete is confined to a single subnet. There is no way to act on IPs matched across subnets (e.g. "deprecate every IP tagged env=legacy", "clear the DNS zone on everything matching hostname regex X") — the Stale-IP report page is the only cross-subnet bulk surface today.

Proposal: a "bulk edit from filter results" affordance at block/space scope (or from a global search result set), reusing the existing `BulkEditAddressesModal` / bulk-delete plumbing. Pairs naturally with server-side search (filed separately).

Issue #517, opened 2026-07-02
ipam: server-side pagination + search for addresses, and IP-table windowing


mzac:
From the 2026-07-02 IPAM review sweep. Three findings converge here — probably the single highest-value IPAM engineering item for scale:

1. **`GET /subnets/{id}/addresses` is unpaginated** — returns every row of a /16 (65k+) with per-row Pydantic serialization (`GET /subnets`, `/blocks` likewise unbounded). `listAddresses` (`frontend/src/lib/api.ts:1637`) takes no params.
2. **The frontend renders all rows with no windowing or memoization** — `filteredAddresses` and the `tableRows` IIFE (`frontend/src/pages/ipam/IPAMPage.tsx:4669, :4720`) are plain expressions recomputed on every keystroke; every row is wrapped in a `ContextMenu` (`:5651+`); the mobile card list is mounted *in addition to* the table (hidden by CSS, `:5336`), doubling element count. The tree got `TREE_GROUP_CAP = 300`; the address table got nothing. A busy /16 locks the tab for seconds.
3. **All filtering is client-side, so there is no cross-subnet address search** — filters only work within one already-downloaded subnet.

Proposal: `?q=&status=&limit=&offset=` (+ tag/hostname/MAC filters) on the addresses endpoint, table virtualization (or a row cap + "show more" like the tree), `useMemo` on the derived lists, and precomputed pool integer bounds (`:4594-4622` re-parses pool boundary strings per row per render). This also unlocks a global "find IP anywhere" search box.

PR #866, opened 2026-08-14
fix(dhcp): adopt-existing retry in the scope modal — end the FortiGate create dead end (#865)

### LINKED ISSUE #865: FortiGate agentless: scope create 409s on adoption conflict with no adopt path — dead end in the scope modal
### Summary

When a DHCP group has a FortiGate (agentless cloud) member and the matching FortiGate interface already carries a DHCP-server object that SpatiumDDI did not create, **creating a scope is a dead end**: the create endpoint pushes to the FortiGate before commit, the push raises the #630 adoption guard (`CloudAdoptionRequired`, 409), and the whole create rolls back. The scope-create modal shows the error text —

> A DHCP server already exists on FortiGate interface 'VLAN_30' that SpatiumDDI did not create. Enable 'adopt existing' to overwrite it with the managed scope, or remove it on the FortiGate first.

— but there is **no "adopt existing" option anywhere in that flow**. The only adopt affordance today is the *Adopt existing & sync* banner on the server detail page (`POST /dhcp/servers/{id}/sync?adopt_existing=true`), which needs an already-persisted scope to push. Chicken-and-egg:

1. Scope create → 409 → rolled back, no scope row.
2. Force Sync → nothing to push (no scope), so the adopt banner can never resolve the conflict.

### Current workaround

Detach the FortiGate server from the group, create the scope (no cloud member → no push), re-attach the server, Force Sync, click *Adopt existing & sync*. Works, but is undiscoverable and clunky.

### Secondary problem

The adoption 409 is indistinguishable from the other scope-create 409 ("A scope for this group+subnet already exists") — both are plain string `detail`s — so the client can't safely decide when offering an adopt-retry is appropriate.

### Proposal

- Accept `adopt_existing: bool = false` on scope **create** and **update** (the activation path can hit the same guard), threaded through `push_scope_upsert` → `push_cloud_scope_upsert`.
- Mark `CloudAdoptionRequired` responses with a machine-readable `X-Adoption-Required: true` header so clients can tell the adoption conflict apart from other 409s.
- Scope modal: try the save without the flag first; on an adoption 409, surface the provider message with an **"Adopt existing & save"** retry that re-submits with `adopt_existing=true` — mirroring the existing Force Sync banner UX.

Adopting keeps the existing FortiOS object in place (same mkey, child-table ids reconciled — no delete/recreate, leases untouched) and records ownership in `scope.provider_refs`, exactly like the sync-path adopt.

### Related

- #630 (adopt guard / provider_refs)
- #625 (small DHCP improvements — this one is bigger than a line item since it needs an API change, hence its own issue)

waza-ari:
Fixes #865.

### Problem

Creating a scope in a group with a FortiGate member pushes to the device **before commit**; when the matching interface already carries a DHCP-server object SpatiumDDI didn't create, the #630 adoption guard raises `CloudAdoptionRequired` (409) and the create **rolls back**. The only adopt affordance — *Adopt existing & sync* on the server detail page — needs a persisted scope to push, so the conflict could never be resolved from the UI (workaround: detach the server from the group, create the scope, re-attach, Force Sync + adopt).

A second problem hid underneath: the adoption 409 was indistinguishable from the duplicate group+subnet 409 (both plain string `detail`s), so a client couldn't safely decide when an adopt-retry is appropriate.

### Change

**Backend**
- `POST /dhcp/subnets/{id}/dhcp-scopes` and `PUT /dhcp/scopes/{id}` accept `?adopt_existing=true`, threaded through `push_scope_upsert` → `push_cloud_scope_upsert` (cloud members only — Windows has no adoption guard, so the flag stops at the seam).
- `CloudAdoptionRequired` responses now carry `X-Adoption-Required: true`, keeping `detail` a plain string (the existing Force Sync banner keeps working unchanged) while giving clients a stable marker.

**Frontend**
- The scope modal always tries the save **without** the flag first. On an adoption 409 (detected via the header — `isAdoptionRequired` in `pages/dhcp/_shared.tsx`) it shows the provider detail in an amber banner with an **"Adopt existing & save"** retry that re-submits with `adopt_existing=true`, mirroring the server page's Force Sync banner. Every other error renders exactly as before.

Adoption semantics are unchanged from #630: the existing FortiOS object is taken over in place (same mkey, child-table ids reconciled — no delete/recreate, leases untouched) and ownership lands in `scope.provider_refs`, so later edits never re-adopt.

### Tests

`backend/tests/test_dhcp_fortigate_agentless_member.py`:
- create → 409 with the header, and the scope row rolled back (no half-persisted state);
- retry with `?adopt_existing=true` → 201, flag observed at the cloud push (`[False, True]`);
- duplicate-scope 409 does **not** carry the header (adopt-retry would be wrong there);
- update path threads the flag too.

Ran serially in the dev container per project guidance: 7/7 in the file, 68 passed across the neighbouring scope/write-through suites, ruff + black + mypy clean, frontend tsc + eslint + prettier clean.

https://claude.ai/code/session_01GtVkdCbknnwHaRJTfhCpHU

### DIFF
--- a/backend/app/api/v1/dhcp/scopes.py
+++ b/backend/app/api/v1/dhcp/scopes.py
@@ -632,8 +632,19 @@ async def _assert_no_overlapping_group_cidr(
     status_code=status.HTTP_201_CREATED,
 )
 async def create_scope(
-    subnet_id: uuid.UUID, body: ScopeCreate, db: DB, user: SuperAdmin
+    subnet_id: uuid.UUID,
+    body: ScopeCreate,
+    db: DB,
+    user: SuperAdmin,
+    adopt_existing: bool = False,
 ) -> ScopeResponse:
+    """Create a scope.
+
+    ``adopt_existing`` (cloud/FortiGate members only, #865) opts in to
+    overwriting a pre-existing provider DHCP object SpatiumDDI never created;
+    without it the pre-commit push 409s (``X-Adoption-Required``) and the
+    create rolls back, so the UI can offer an adopt-and-retry.
+    """
     subnet = await db.get(Subnet, subnet_id)
     if subnet is None:
         raise HTTPException(status_code=404, detail="Subnet not found")
@@ -731,7 +742,7 @@ async def create_scope(
         ) from exc
     # Push to every Windows DHCP member of the group BEFORE commit so a
     # WinRM failure rolls the DB row back.
-    await push_scope_upsert(db, scope)
+    await push_scope_upsert(db, scope, adopt_existing=adopt_existing)
     collect_wake(dhcp_group_channel(group_id))
     write_audit(
         db,
@@ -757,8 +768,15 @@ async def get_scope(scope_id: uuid.UUID, db: DB, _: CurrentUser) -> ScopeRespons
 
 @router.put("/scopes/{scope_id}", response_model=ScopeResponse)
 async def update_scope(
-    scope_id: uuid.UUID, body: ScopeUpdate, db: DB, user: SuperAdmin
+    scope_id: uuid.UUID,
+    body: ScopeUpdate,
+    db: DB,
+    user: SuperAdmin,
+    adopt_existing: bool = False,
 ) -> ScopeResponse:
+    # ``adopt_existing``: same opt-in as create (#865) — an edit of a scope
+    # whose interface carries a foreign provider object hits the same
+    # adoption guard on the pre-commit push.
     scope = await db.get(DHCPScope, scope_id)
     if scope is None:
         raise HTTPException(status_code=404, detail="Scope not found")
@@ -827,7 +845,7 @@ async def update_scope(
     for k, v in changes.items():
         setattr(scope, k, v)
     await db.flush()
-    await push_scope_upsert(db, scope)
+    await push_scope_upsert(db, scope, adopt_existing=adopt_existing)
     collect_wake(dhcp_group_channel(scope.group_id))
     write_audit(
         db,
--- a/backend/app/main.py
+++ b/backend/app/main.py
@@ -780,10 +780,14 @@ def create_app() -> FastAPI:
         allow_headers=["*"],
         # Custom response headers the browser must be allowed to read via
         # fetch(). ``X-Total-Count`` backs server-side pagination on the IPAM
-        # address list + cross-subnet search (issues #517 / #520). Without
-        # this the header is present on the wire but the JS layer can't read
-        # it under CORS.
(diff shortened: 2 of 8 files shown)


mzac:
Reviewed — the design is right. The flag stopping at the Windows seam, always attempting the save without it first, and keeping `detail` a plain string so the existing Force Sync banner is untouched are all the correct calls. `push_cloud_scope_upsert` already took the kwarg, so the threading is genuinely additive.

**Verified locally:** ruff + black clean; mypy identical to the `main` baseline (120 pre-existing errors, none in touched files); frontend build + eslint + prettier clean; `test_dhcp_fortigate_agentless_member.py` 7/7 and 54 more across the neighbouring scope/write-through suites.

### One gap that disables the fix under CORS

`X-Adoption-Required` is set on the response but never added to the CORS middleware's `expose_headers`, which lists only `X-Total-Count` (`backend/app/main.py:786`). For a cross-origin frontend — an explicitly supported deployment, see the `cors_origins` note in `config.py` — the browser refuses to expose the header to `fetch()`, so `isAdoptionRequired()` reads `undefined`, the amber banner never renders, and the flow degrades to the generic 409: exactly the dead end this PR fixes. It's the same trap the comment directly above that line already documents for `X-Total-Count` (#517 / #520).

```python
expose_headers=["X-Total-Count", "X-Adoption-Required"],
```

### Minor

`adoptConflict` isn't cleared in `onMutate`, so a stale conflict banner stays on screen while a fresh save is in flight. Cosmetic.

### On merging

This is a fork PR, and Code Quality / CodeQL `dynamic` runs are never created for fork PRs in this repo — the required `code_quality` rule sits pending forever, so it can't merge as-is. I've re-pushed your commit (authorship preserved) in-repo as `issue-865` with the `expose_headers` fix on top as a separate commit, so the checks can run. Thanks for the fix and the clear write-up.

mzac:
Update, and a correction to my earlier comment.

We've fixed the underlying problem rather than working around it: the `code-quality-gate` and Copilot-review rulesets now carry an admin bypass scoped to pull requests, so a fork PR whose Copilot/code-quality gate can never run is no longer unmergeable. **This PR merges as yours** — please disregard what I said earlier about a replacement branch, that plan is dropped and the `issue-865` branch will be deleted.

I've pushed the one fix from my review directly onto this branch instead (`cd43688`): `X-Adoption-Required` added to the CORS middleware's `expose_headers`. Everything else here is your work, unchanged.

Once the workflow runs are approved and green this is ready to merge. Thanks for the fix — and for being the PR that got the fork-contribution path fixed for everyone after you.

Issue #631, opened 2026-07-12
dhcp: allow static reservations inside dynamic pool ranges — the in-pool 409 has no reason, and looks like it breaks Windows DHCP


mzac:
## Summary

We currently refuse to create a static DHCP reservation whose IP falls inside a
`dynamic` pool. Reported on Discord by **Wasabi**:

> One comment - currently, we block creation of static DHCP allocs within a dynamic
> range - e.g. range 10.10.10.100-150, trying to add a static reservation for
> 10.10.10.120 is blocked. I wonder why that is - Kea perfectly supports this, so
> does Fortigate. In fact its quite common.
>
> Was there a specific reason you prevented that? If not I propose to change it.

**There was no specific reason.** The check landed in `601f4dee` (2026-04-15), the
very first DHCP backend commit — one-line commit message, no design note, no
issue behind it, and to this day **no test asserts it**. It's an initial-implementation
assumption, not a considered decision.

Wasabi is right and it's worse than "merely unnecessary": in-pool reservations are
the *standard idiom* on all three drivers we ship, the rule's own suggested workaround
is a dead end on every driver, three of our own code paths already bypass it, and it
looks like it silently breaks static reservations on the Windows DHCP driver
entirely. **Proposal: drop it.**

## 1. No driver needs the rule

| Driver | In-pool reservation today | Notes |
|---|---|---|
| **Kea** | ✅ Honored, zero config change | We never set `reservations-out-of-pool` anywhere (grepped repo-wide, incl. the baked `agent/dhcp/images/kea/kea-dhcp4.conf`). So Kea defaults apply — `reservations-out-of-pool: false` — which is *precisely* the mode where Kea runs the in-pool check and won't hand a reserved address to another client. Pools (`kea.py:208`) and reservations (`kea.py:169-191`) are emitted as independent sibling keys with no cross-check. |
| **FortiGate** | ✅ Honored | Statics → `reserved-address` on `system.dhcp.server` (`fortigate.py:364-384`); `ip-range` is built independently from dynamic pools (`fortigate.py:321-336`). No cross-check. Pinning a MAC inside the range is the normal FortiOS idiom. |
| **Windows** | ⚠️ **Requires in-pool** | Windows keys reservations to the scope's StartRange–EndRange and rejects a reservation outside it. See §3. |

## 2. The rule's own advice is a dead end

The IPAM error tells the operator to go use a static reservation
(`ipam/router.py:6586-6598`, **422**):

[...]

Issue #516, opened 2026-07-02
ipam: frontend polish grab-bag (error handling, dead props, convention violations, small v6/status gaps)


mzac:
Found by the 2026-07-02 IPAM review sweep. Individually small; grouped as one polish pass. All in `frontend/src/pages/ipam/` unless noted.

1. **Static-DHCP chained create misreports.** `AddAddressModal` chains `dhcpApi.createStatic` inside the same `mutationFn` (`IPAMPage.tsx:3181-3193`); if the IP create succeeds but the reservation 4xx's, the modal shows "Failed to allocate address" although the row exists — re-submitting collides. Separate the error reporting.
2. **`IPDetailModal` zone fields always render "—".** `zoneNameById` is a dead prop no caller passes (`IPDetailModal.tsx:112, 145, 341-358`) — reads as "no zone" for IPs that have one. Resolve names in the modal or drop the fields.
3. **Silent failures:** `EditSpaceModal.saveMutation` (`:10368-10387`) and `SubnetDetail`'s delete/purge/restore mutations (`:4815-4854`) have no `onError` — a 403/409 leaves the modal open with zero feedback.
4. **`BulkEditSubnetsModal` is a raw fixed div** (`:13829`), not the shared draggable `Modal` — no drag, no Esc, wrong backdrop; violates the project modal convention. Its "leave blank to keep" semantics also make bulk-clearing impossible.
5. **`DnsSyncModal` mutates on open** — reverse-zone backfill fires in a mount `useEffect` (`:6465-6482`) before the operator approves anything; surprising for a preview-then-commit surface and awkward with the #62 approvals story.
6. **Status dropdowns omit `available`/`discovered`** (`IP_STATUS_OPTIONS:2819`, `ADDRESS_STATUSES:8041`; bulk edit offers `available` at `:9453`) — flipping one IP back to `available` requires bulk-editing a single row, and editing a `discovered` row renders a blank select.
7. **Skipped-rows feedback is a 1.2 s flash** (`:9397-9402`, `:9831-9836`) — the only report of which rows were skipped disappears before it can be read.
8. **`cidrSize` fallback wrong for IPv6** — `Math.pow(2, 32 - prefix)` on a /48 yields 2^-16 (`:12795-12798`), feeding garbage into `UsedIps` when `total_ips` is null.
9. **`SyncMenu` `isPending` hardwired `false`** (`:4916`) — spinner/disable plumbing is dead.

Issue #472, opened 2026-07-02
DHCP: enable creating static reservations from the Static Assignments tab (wire up dead CreateStaticAssignmentModal)


mzac:
## Problem

There is no way to **create** a DHCP static reservation from the DHCP UI. The DHCP server group's **"Static Assignments"** tab (`ServerPoolsOrStaticsTab`, `frontend/src/pages/dhcp/DHCPPage.tsx`) is strictly **read-only** — a sortable table plus a copy-only right-click menu, with a bare "No static assignments yet." empty state and no add/edit/delete affordance.

The purpose-built `CreateStaticAssignmentModal` / `EditStaticAssignmentModal` (`frontend/src/pages/dhcp/CreateStaticAssignmentModal.tsx`) already exist but are **dead code — imported nowhere**.

Today the *only* working path to create a Kea reservation is IPAM → Allocate IP → status `static_dhcp` + scope + MAC (`IPAMPage.tsx` `AddAddressModal`), which is non-obvious and silently no-ops on a blank MAC / edit-of-existing-IP / non-superadmin. This is what a user hit (Discord report) and what the docs fix in #471 documents.

## Proposed change (frontend)

1. **Wire `CreateStaticAssignmentModal` into the Static Assignments tab** — add an "Add static assignment" button (follow the `ClientClassesTab` `setShowCreate` pattern in the same file). Since the tab is group-centric across multiple scopes, the create modal needs a **scope picker** (which scope to add the reservation to).
2. **Wire `EditStaticAssignmentModal`** — add Edit + Delete to the per-row context menu (keep the existing copy actions).
3. **Deep-link the IPAM hint** — the amber "No DHCP scope exists for this subnet. Create one from the DHCP Pools tab first." message in `AddAddressModal` (`IPAMPage.tsx`) should be a button that opens `CreateScopeModal` for the subnet inline.
4. Invalidate the relevant React Query keys (`dhcp-statics`, per-scope) so the table + IPAM `static_dhcp` picker refresh.

## Notes / guardrails

- Backend create/update/delete endpoints already exist (`POST/PUT/DELETE /dhcp/scopes/{scope_id}/statics`, `backend/app/api/v1/dhcp/statics.py`) and are **superadmin-gated** — the UI should reflect that (hide/disable for non-superadmins).
- Docs for the current (IPAM-only) path are handled in #471; this issue is the code follow-up.

Reported via Discord; split out of #471 (docs).

PR #868, opened 2026-08-14
fix(ipam): edit-address modal can now pin a DHCP reservation when flipping a row to static_dhcp (#867)

### LINKED ISSUE #867: ipam: editing an allocation to status=static_dhcp is a silent no-op — no scope picker, no reservation, and #516's own "edit the row to retry" advice is impossible
## Summary

Editing an IP allocation and setting **Status = `static_dhcp`** silently does nothing on the DHCP side: the edit modal has **no DHCP scope picker** and never creates a `DHCPStaticAssignment`. The create modal *does* both (scope select + chained `dhcpApi.createStatic`). So an operator who allocates an IP first and wants to pin the reservation later — e.g. because the scope was created *after* the allocation — gets a row that *looks* reserved while the DHCP server keeps leasing the address.

## Why this is a trap, not just a missing convenience

1. **The save succeeds and looks correct.** `PUT /ipam/addresses/{id}` happily accepts `status="static_dhcp"` (it even enforces the MAC requirement, `ipam/router.py` `update_address`), but neither the backend nor the edit modal creates a reservation. The subnet view then renders the row as a static reservation. Nothing tells the DHCP server to stop handing the IP out — the exact duplicate-IP risk #631 called out.

2. **Our own error message points at this dead end.** The create modal's #516 partial-failure handler (address created, chained reservation failed) tells the operator:

   > "The address row was created — **close this and edit the row to retry the reservation**."

   That retry path does not exist. The edit modal cannot create a reservation.

3. **The stated workaround is undiscoverable.** PR #634 (which implemented #631) scoped this out with "the operator is reminded to also pin one on the scope" — and creating the reservation from the DHCP side does work (`upsert_ipam_for_static` finds the existing IPAM row by subnet+IP and links it, no duplicate). But nothing in the IPAM edit flow says so, and the natural conclusion (reported by an operator, which is how this was found) is that you must **delete and re-create** the allocation to reach the scope picker.

## Not a conscious design decision — a consciously *deferred* one, minus the UI half

#631's open question explicitly documented the one-way mirror ("setting an IPAM row to `status=static_dhcp` does **not** create a `DHCPStaticAssignment` — the mirror only flows reservation → IPAM") and deferred the backend auto-create as a follow-up with its own blast radius. That deferral is fine. But the *edit-modal* asymmetry was never part of that decision — the create modal's client-side chaining pattern (#516) simply was never mirrored to the edit modal, and the #516 error message assumes it was.

## Proposed fix (frontend parity, no backend behavior change)

In `EditAddressModal` (`frontend/src/pages/ipam/IPAMPage.tsx`):

- When **Status = `static_dhcp`** and the row has **no linked reservation** (`address.static_assignment_id == null`): show the same DHCP scope picker as the create modal (scopes via `dhcpApi.listScopesBySubnet`, amber empty-state with the inline **Create a scope** link from #472), and on Save chain `dhcpApi.createStatic(scopeId, …)` after `updateAddress` succeeds — with the same partial-failure handling as #516 (row saved, reservation failed → keep modal open, explain, don't re-submit the row).
- When a reservation **is** already linked: show an informational line instead of the picker (reservation is the source of truth for hostname/MAC on the DHCP side).
- When moving **off** `static_dhcp` while a reservation is linked: an advisory note that the DHCP reservation itself is not removed by this edit (remove it on the DHCP side to actually free the address for the pool).

**Explicitly out of scope** (belongs to #631's deferred follow-up, server-side): making the backend create/remove the reservation on `status` transitions so the pure REST path gets the same semantics. The client-side chaining here mirrors the existing, established create-modal pattern and at minimum removes the silent-trap + restores the #516 retry path.

*File refs against `main` @ `1e9493a9`.*

waza-ari:
Fixes #867.

## Problem

Editing an IP allocation to **Status = `static_dhcp`** was a silent no-op on the DHCP side: `EditAddressModal` had no DHCP scope picker and never created a `DHCPStaticAssignment` — the row *looks* reserved in the subnet view while the DHCP server keeps leasing the address (the exact duplicate-IP risk #631 documented). The create modal has had the scope picker + chained `createStatic` since #516, and its partial-failure message even tells the operator to *"close this and edit the row to retry the reservation"* — a retry path that didn't exist until now.

## Change (frontend-only, mirrors the established create-modal pattern)

In `EditAddressModal`:

- **Scope picker** when `status === "static_dhcp"` and the row has no linked reservation (`static_assignment_id == null`): scopes via `dhcpApi.listScopesBySubnet` (query gated on need), first scope auto-selected, amber empty-state with the #472 inline **Create a scope** escape hatch — all identical to `AddAddressModal`.
- **Chained reservation create** on Save: after `updateAddress` succeeds, `dhcpApi.createStatic(scopeId, {ip, mac, hostname, description})` with the #516 partial-failure contract. Unlike create, the row update is idempotent, so on a failed reservation the modal stays open and pressing Save again retries it (the error says so).
- **Linked-row note**: when a reservation is already linked, an informational line replaces the picker (the reservation is the source of truth for hostname/MAC via `upsert_ipam_for_static`).
- **Transition-off advisory**: moving a row off `static_dhcp` while a reservation is linked warns that the DHCP reservation itself is not removed by the edit.
- Invalidates `["dhcp-statics", scopeId]` so the DHCP statics tab reflects the new reservation.

**Explicitly out of scope** (per #631's deferred open question): server-side auto-create/remove of the reservation on `status` transitions, so the pure REST path gets the same semantics. This PR removes the UI trap and restores the promised #516 retry path without changing backend behavior.

## Verification

- `npx prettier --check`, `npx eslint`, and `npm run build` clean on the touched file (chunk-size warning pre-existing).
- Manual flow: allocate → later create scope → edit row → pick `static_dhcp` → scope picker appears → Save creates the reservation and links the row (`static_assignment_id` set by `upsert_ipam_for_static`); re-opening the modal shows the linked-reservation note instead of the picker.

https://claude.ai/code/session_01GtVkdCbknnwHaRJTfhCpHU

### DIFF
--- a/frontend/src/pages/ipam/IPAMPage.tsx
+++ b/frontend/src/pages/ipam/IPAMPage.tsx
@@ -8681,6 +8681,27 @@ function EditAddressModal({
   const [pendingWarnings, setPendingWarnings] = useState<
     CollisionWarning[] | null
   >(null);
+  // #867 — parity with AddAddressModal: flipping an existing row to
+  // ``static_dhcp`` must be able to pin the reservation on a scope instead of
+  // silently saving a row the DHCP server keeps leasing. Only surfaced when no
+  // reservation is linked yet — a linked row is owned by the DHCP side
+  // (``upsert_ipam_for_static`` is the source of truth for hostname/MAC).
+  const [dhcpScopeId, setDhcpScopeId] = useState<string>("");
+  const [showCreateScope, setShowCreateScope] = useState(false);
+  const needsDhcpScope =
+    status === "static_dhcp" && !address.static_assignment_id;
+
+  const { data: dhcpScopes = [] } = useQuery({
+    queryKey: ["dhcp-scopes-subnet", address.subnet_id],
+    queryFn: () => dhcpApi.listScopesBySubnet(address.subnet_id),
+    enabled: needsDhcpScope,
+  });
+
+  useEffect(() => {
+    if (needsDhcpScope && !dhcpScopeId && dhcpScopes.length > 0) {
+      setDhcpScopeId(dhcpScopes[0].id);
+    }
+  }, [needsDhcpScope, dhcpScopes.length]); // eslint-disable-line react-hooks/exhaustive-deps
 
   const { data: cfDefs = [] } = useQuery({
     queryKey: ["custom-fields", "ip_address"],
@@ -8779,7 +8800,7 @@ function EditAddressModal({
       : null;
 
   const mutation = useMutation({
-    mutationFn: (force: boolean) => {
+    mutationFn: async (force: boolean) => {
       // Empty string on role → null (clears the field).
       // Empty string on reservedUntil → null when status=reserved
       // (= no TTL); when status moved off reserved we still send
@@ -8788,7 +8809,7 @@ function EditAddressModal({
         status === "reserved" && reservedUntil
           ? new Date(reservedUntil).toISOString()
           : null;
-      return ipamApi.updateAddress(address.id, {
+      const updated = await ipamApi.updateAddress(address.id, {
         // Send explicit null/"" so an operator can *clear* a field — the old
         // ``value || undefined`` was dropped by axios and ignored by the
         // backend's exclude_unset, making a clear a silent no-op (#502).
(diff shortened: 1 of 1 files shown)


mzac:
Reviewed — clean, and a faithful mirror of `AddAddressModal` rather than a second pattern. Keeping the server-side auto-create/remove on `status` transitions out of scope (per #631's deferred question) is the right call for this PR.

**Verified locally:** `npm run build`, eslint and prettier all clean. I also checked the details that usually drift in this kind of change, and they're all correct:

- `["dhcp-statics", scopeId]` matches the key the DHCP statics tab actually reads (`DHCPPage.tsx:1143`), so the invalidation lands;
- `["dhcp-scopes-subnet", subnetId]` is the same key `AddAddressModal` uses, so the two share one cache entry instead of double-fetching;
- `CreateScopeModal`'s props line up and it was already imported in `IPAMPage.tsx`.

### One thing worth deciding (not a blocker, and not a regression)

Save still succeeds silently with no reservation when the MAC is empty or the subnet has no scope — the guard is `needsDhcpScope && dhcpScopeId && macAddress`, and only an amber advisory marks either case. That's exact parity with the create modal's `&& mac` guard, so this PR doesn't make anything worse, but it does leave the "row looks reserved, DHCP keeps leasing" trap open in those two paths. Gating the Save button on it — in both modals, so they stay in step — would close it. Happy for that to be a follow-up rather than scope creep here.

### On merging

This is a fork PR, and Code Quality / CodeQL `dynamic` runs are never created for fork PRs in this repo — the required `code_quality` rule sits pending forever, so it can't merge as-is. I've re-pushed your commit (authorship preserved) in-repo as `issue-867` so the checks can run. Thanks for catching this one — the #516 message promising a retry path that didn't exist was a good find.

Issue #805, opened 2026-08-02
Nightly build pipeline for main — publish testable images + ISO between releases


mzac:
## Problem

There is no way to test `main` without cutting a CalVer release. Every published
image and ISO comes from `release.yml`, which fires only on a
`YYYY.MM.DD-N` tag. So:

- Between releases, the only way to run current `main` is to build locally
  (`make build`, or the multi-step appliance ISO recipe) — slow, and it
  exercises a different path from what CI actually publishes.
- Integration bugs that only appear in a **published, multi-arch, non-dev-stage**
  image surface for the first time during release prep, which is the worst
  moment to find them. [#732](https://github.com/spatiumddi/spatiumddi/issues/732)
  (the api image shipping the `dev` stage) is exactly this class: it survived
  because nothing built the release image except a release.
- The weekly Trivy sweep (`trivy-scheduled.yml`) scans *published* images, so a
  base-image CVE introduced by a Dockerfile change sits undetected until the
  next release publishes something for it to scan.

A nightly build of `main` closes all three.

## Proposal

New `.github/workflows/nightly.yml` — scheduled build of `main` that publishes
the same artifact set `release.yml` does, under nightly-only tags that can never
be confused with a release.

### Retention: one pointer, plus a short bisect ring

The instinct not to accumulate junk is right, but "keep exactly one" needs to
split by artifact, because the two have different failure modes.

**Container images — one mutable pointer + a 7-day ring.**

| Tag | Meaning | Retention |
|---|---|---|
| `:nightly` | Always the most recent nightly | Mutable, overwritten each run |
| `:nightly-YYYYMMDD` | Immutable snapshot of that night | Prune to the newest 7 |

Keeping *literally* one image would mean that when someone reports "nightly
broke X", the tag has already rolled and there is nothing to compare against —
no bisect, and no fallback if a nightly is broken. Seven dated tags cost
nothing in clutter (they prune in the same workflow run) and turn "it broke
sometime this week" into a bounded search. Anyone who just wants to run
nightly pulls `:nightly` and never sees the rest.

GHCR storage is free for public packages, so this is about the package version
list staying navigable, not cost.

**Appliance ISO + slot image — genuinely keep exactly one.**

These are ~1.5 GB each (3 GB per build), and `build-appliance-iso` has a
60-minute timeout and depends on all ten image builds. Here the instinct
applies without qualification: publish to a **single `nightly` pre-release**
whose assets are replaced in place, so the download URL is stable and nothing
[...]

PR #970, opened 2026-09-04
ci(nightly): gate the DST twin crons on which cron fired, not the run-time clock

amoona6:
## Summary

**No scheduled nightly has built since 2026-08-09.** Every `schedule` run of `nightly.yml` since then completes "success" in 10–18 s having done nothing; the only nightly in the window is the 08-08 manual dispatch (`nightly-2026.08.08`, built-from `08d81ad`), while `main` has moved 30+ commits. The `Decide` gate tests the Eastern wall clock **when the job runs** (`TZ=America/New_York date +%H` must be `02`), but GitHub starts these crons late — every night, by a lot — so neither twin has ever woken inside 02:xx and both skip.

Measured start times of the twins (should be 02:23 EDT):

| night | 06:23 UTC twin started | 07:23 UTC twin started | late by |
|---|---|---|---|
| 08-09 | 03:26 EDT | 04:05 EDT | 1 h |
| 08-28 | 14:49 EDT | 15:29 EDT | 12–13 h |
| 09-01 | 07:46 EDT | 08:37 EDT | 5 h |
| 09-04 | 07:27 EDT | 08:10 EDT | 5 h |

(`gh run list --workflow nightly.yml`, 56 scheduled runs since 08-09, all 10–18 s, all skipped with `→ HH:MM EDT is not the 02:xx Eastern slot; skipping (DST twin cron)`.)

This PR keeps the twin-cron design and fixes the gate: decide from **which cron fired** (`github.event.schedule`, the cron string, immune to scheduler delay), reading that cron's scheduled UTC instant on the Eastern clock and keeping the twin whose slot is 02:xx. A missing or malformed cron string now fails the gate (so `finalize` opens the tracking issue) instead of quietly skipping — a quiet skip is exactly how this went unnoticed for four weeks. `workflow_dispatch` behaviour is unchanged.

Behaviour on the two DST nights, for the record: fall-back → the 07:23 twin reads 02:23 EST, one build; spring-forward → no Eastern 02:xx exists that night, neither twin builds, the next night covers it (same as the design's stated intent).

Considered and not taken: dropping the time gate and letting the second twin skip on "today's pre-release already exists". Delay-immune too, and it would even cover a dropped first twin, but it changes which slot builds in winter (01:23 EST) and re-runs a failed build an hour later; the one-line-in-spirit fix above preserves the 02:23-year-round intent exactly.

## Area

- [x] Deployment (Compose / K8s / Appliance) — CI, `nightly.yml` only

## Test plan

- [x] **faketime matrix of the exact gate block** (GNU date 9.4 + tzdata, the same coreutils as `ubuntu-latest`): 15 cases — both twins delayed 5 h and 13 h in EDT, on time in EDT and EST, delayed in EST, a delay past UTC midnight, the 2026-11-01 fall-back and 2027-03-14 spring-forward nights. All decide as documented above; the previous gate skips every delayed case. Plus: empty and garbage `github.event.schedule` exit 1 with a `::error`, and `workflow_dispatch` passes through.

  <details><summary>Matrix output on <code>ubuntu-latest</code> (spatiumddi-qa run 33902702507, all 15 match, empty schedule → exit 1, dispatch passes through)</summary>

  ```
  CASE                               SCHEDULE     NOW (UTC)            WANT   GOT
  EDT delayed 5h, twin1              23 6 * * *   2026-09-04 11:27:00  build  build
  EDT delayed 5h, twin2              23 7 * * *   2026-09-04 12:10:00  skip   skip
  EDT delayed 13h, twin1             23 6 * * *   2026-08-28 18:49:00  build  build
  EDT delayed 13h, twin2             23 7 * * *   2026-08-28 19:29:00  skip   skip
  EDT on time, twin1                 23 6 * * *   2026-09-04 06:23:30  build  build
  EDT on time, twin2                 23 7 * * *   2026-09-04 07:23:30  skip   skip
  EDT delayed past UTC midnight      23 6 * * *   2026-09-05 00:30:00  build  build
  EST delayed, twin1                 23 6 * * *   2026-12-04 07:00:00  skip   skip
  EST delayed, twin2                 23 7 * * *   2026-12-04 08:30:00  build  build
  EST on time, twin2                 23 7 * * *   2026-12-04 07:23:30  build  build
  spring-fwd 2027-03-14 twin1 early  23 6 * * *   2027-03-14 06:50:00  skip   skip
  spring-fwd 2027-03-14 twin1 late   23 6 * * *   2027-03-14 07:30:00  skip   skip
  spring-fwd 2027-03-14 twin2        23 7 * * *   2027-03-14 08:00:00  skip   skip
  fall-back 2026-11-01 twin1         23 6 * * *   2026-11-01 06:40:00  skip   skip
  fall-back 2026-11-01 twin2         23 7 * * *   2026-11-01 07:40:00  build  build
  ```
  The previous gate, evaluated at the same instants, builds only where the delayed start happens to land inside 02:xx (four rows: the two on-time slots, plus 02:00 EST and 02:40 EST by luck) and skips every other row — including every delayed EDT night, which is every night since 08-09.
  </details>

- [x] **Real `schedule` event** in the private `spatiumddi-qa` sandbox, on a probe workflow that runs the gate block extracted byte-for-byte from this PR's `nightly.yml`: the `23 18 * * *` cron was started by GitHub **2 h 26 min late** (20:48 UTC) and the gate still decided from the cron string — `is the 14:23 EDT slot … skipping` — see the comment below for the log. The real `23 6`/`23 7` twins run on the same probe tonight (expected build/skip) and will be posted as a comment.
- [x] `actionlint` on the changed file: no new findings (one pre-existing SC2129 style note on the original `$GITHUB_OUTPUT` echoes).
- [ ] The first real proof is tomorrow's 06:23 UTC twin on `main` after merge — it should log `→ cron '23 6 * * *' is the 02:23 EDT slot; run started HH:MM EDT` and go on to build.

Downstream context: ddi-pg's full-lane QA (cluster_gate/api_cluster/gate_load/ha/ha_upgrade) walks the nightly's ISO + slot image, so it has had nothing new to walk since 08-08; its wall raises a "nightly lane STALE" banner for exactly this.

## Related issues

Refs #849 (the twin-cron rewrite this fixes), #805.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -55,8 +55,12 @@ on:
     # so the nightly must run AFTER their late pushes, DST included. Cron
     # is UTC-only, so schedule both candidate UTC hours (06:23 = 02:23
     # EDT, 07:23 = 02:23 EST) and let the gate job keep exactly the one
-    # where the Eastern wall clock actually reads 02 — the other fires
-    # and immediately skips. Off-the-hour minute for the same reason
+    # whose SCHEDULED slot reads 02 on the Eastern clock — the other fires
+    # and immediately skips. The gate decides from the cron string GitHub
+    # hands it (github.event.schedule), never from the clock at run time:
+    # GitHub starts these runs 40 min to 13 h late (measured, every
+    # night), so the run-time wall clock says nothing about which slot a
+    # run belongs to. Off-the-hour minute for the same reason
     # prune-release-assets.yml uses one: the top-of-hour cron stampede
     # delays runner scheduling on GitHub's shared infrastructure.
     - cron: "23 6 * * *"
@@ -114,18 +118,49 @@ jobs:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           FORCE: ${{ github.event.inputs.force || 'false' }}
           EVENT_NAME: ${{ github.event_name }}
+          # The cron string that fired ("23 6 * * *"); empty on dispatch.
+          SCHEDULE: ${{ github.event.schedule }}
         run: |
           set -euo pipefail
 
-          # ── Eastern-wall-clock gate ──────────────────────────────────
+          # ── Which twin fired? ────────────────────────────────────────
           # Two UTC crons cover 02:23 America/New_York across DST (see the
-          # trigger comment). Keep the firing where Eastern actually reads
-          # 02; the other is a no-op. Manual dispatch always proceeds.
-          ET_HOUR=$(TZ=America/New_York date +%H)
-          if [ "$EVENT_NAME" = "schedule" ] && [ "$ET_HOUR" != "02" ]; then
-            echo "should_build=false" >> "$GITHUB_OUTPUT"
-            echo "→ $(TZ=America/New_York date '+%H:%M %Z') is not the 02:xx Eastern slot; skipping (DST twin cron)"
-            exit 0
+          # trigger comment). Keep the twin whose SCHEDULED slot reads 02
+          # on the Eastern clock; the other is a no-op. Manual dispatch
+          # always proceeds.
+          #
+          # The slot comes from the cron string that fired, NOT from the
+          # clock when this step runs. GitHub starts scheduled runs late —
+          # 40 min to 13 h on this repo, every night since the twin crons
+          # landed — and the previous form of this gate compared the
+          # run-time Eastern hour to 02, so both twins always woke outside
+          # 02:xx and both skipped: no scheduled nightly built between
+          # 2026-08-09 and this change (the only nightly in that window
+          # was a manual dispatch). The cron string is immune to the delay.
+          #
+          # DST nights: on the fall-back night the 06:23 twin reads 01:23
+          # EST and the 07:23 twin 02:23 EST — one build. On the
+          # spring-forward night no Eastern 02:xx exists (01:59 → 03:00),
+          # so neither twin builds; the next night covers it.
+          if [ "$EVENT_NAME" = "schedule" ]; then
+            # "23 6 * * *" → "6:23", today's 06:23 UTC read on the Eastern
+            # clock. A missing or odd cron string is a hard failure, never
+            # a skip — a failed gate is what finalize turns into the
+            # tracking issue, whereas a quiet skip is exactly how the
+            # previous gate went unnoticed for four weeks.
+            CRON_UTC=$(printf '%s' "$SCHEDULE" | awk '{print $2 ":" $1}')
+            if [[ ! "$CRON_UTC" =~ ^[0-9]{1,2}:[0-9]{1,2}$ ]]; then
+              echo "::error::github.event.schedule is '${SCHEDULE}' — cannot tell which twin fired"
+              exit 1
+            fi
+            SLOT_ET=$(TZ=America/New_York \
+                      date -d "$(date -u +%F) ${CRON_UTC} UTC" '+%H:%M %Z')
+            if [ "${SLOT_ET%%:*}" != "02" ]; then
+              echo "should_build=false" >> "$GITHUB_OUTPUT"
+              echo "→ cron '${SCHEDULE}' is the ${SLOT_ET} slot, not 02:xx Eastern; skipping (DST twin cron)"
+              exit 0
+            fi
+            echo "→ cron '${SCHEDULE}' is the ${SLOT_ET} slot; run started $(TZ=America/New_York date '+%H:%M %Z')"
           fi
 
           # ONE source of truth for which images exist. Both the build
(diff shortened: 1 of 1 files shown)


amoona6:
**Live `schedule` event proof, part 1 — the gate block from this PR on a real cron, under a real delay.**

Probe in the private `spatiumddi-qa` sandbox (workflow runs the block extracted byte-for-byte from this PR's `nightly.yml`, plain `run:` steps, no build/publish jobs). Its `23 18 * * *` cron was started by GitHub at **20:48:58 UTC — 2 h 26 min late** (the same class of delay that has been skipping the real nightly), and the gate decided correctly from the cron string regardless of the clock:

```
event=schedule schedule='23 18 * * *'
run_id=33918099761   step now: 2026-09-04T20:49:04Z = 16:49 EDT
→ cron '23 18 * * *' is the 14:23 EDT slot, not 02:xx Eastern; skipping (DST twin cron)
decision: skip   expected: skip
```

So `github.event.schedule` reaches the step as the plain cron string, the parse + `date -d` work on `ubuntu-latest`'s coreutils/tzdata, and the decision is the slot's, not the start time's. The previous gate on this same run would have printed `→ 16:49 EDT is not the 02:xx Eastern slot; skipping` — identical outcome here only because this cron *is* a skip case.

Still to come, posted here as they land: the probe's real `23 6 * * *` / `23 7 * * *` twins tonight — expected `build` (02:23 EDT slot) and `skip` (03:23 EDT slot) respectively, whatever time GitHub actually starts them.

mzac:
Reviewed, and I pushed one commit to the branch: `c1651a70`. The diagnosis and the fix are right, the twin-cron design is worth keeping, and I verified the three load-bearing claims rather than taking them:

- `actionlint` on the changed file reports only the pre-existing SC2129 style note, as stated.
- The hard-failure path really does reach the tracking issue. `finalize` is `always() && needs.gate.result != 'skipped' && needs.gate.outputs.should_build != 'false'`, and a failed gate emits no outputs, so the empty string passes that last test and `needs.gate.result == 'failure'` is the first condition on the issue step. The comment above `finalize` already anticipates exactly this.
- Every one of the 15 matrix rows decides as documented. I re-ran them from the extracted block, not from a re-implementation.

## The one finding

The gate now takes the slot from the cron string, which is delay-immune. It then resolves that against `$(date -u +%F)`, which is still the clock at run time. The two disagree when a delay crosses UTC midnight, and that only changes the answer when the day it lands on has a different UTC offset. So, a DST-change night.

| case | cron | started (UTC) | pre-patch | true slot |
|---|---|---|---|---|
| 10-31 twin1, +18 h | `23 6 * * *` | 11-01 00:30 | skip | 02:23 EDT, build |
| 10-31 twin2, +17 h | `23 7 * * *` | 11-01 00:30 | build | 03:23 EDT, skip |
| 03-13 twin2, +17 h | `23 7 * * *` | 03-14 00:30 | skip | 02:23 EST, build |

Rows 1 and 2 together still yield one build, by luck, on the wrong twin. The bad combination is row 1 with its sibling running on time: twin1 crosses midnight and skips, twin2 reads 03:23 EDT and skips, and the night builds nothing silently. That is this PR's own failure mode, at twice a year instead of nightly. The spring-forward variant loses the night's only build the same way.

It needs about 17 h of delay against the 13 h you measured, so it is reachable rather than theoretical.

## The fix

The slot is the most recent occurrence of that cron time at or before now. GitHub starts runs late but never early, so stepping back a day when today's instant is still in the future names the actual slot:

```bash
SLOT_UTC=$(date -u -d "$(date -u +%F) ${CRON_UTC} UTC" +%s)
if [ "$SLOT_UTC" -gt "$(date -u +%s)" ]; then
  SLOT_UTC=$((SLOT_UTC - 86400))
fi
SLOT_ET=$(TZ=America/New_York date -d "@$SLOT_UTC" '+%H:%M %Z')
```

Written as an `if` rather than `[ … ] && SLOT_UTC=…` on purpose: the `&&` form returns 1 when the condition is false, which under `set -e` is a footgun waiting for the first quiet night.

**Verification.** I extracted the gate block byte-for-byte from the workflow YAML and ran it under `faketime` on Debian with GNU date and tzdata, the same coreutils as `ubuntu-latest`. All 15 of your rows unchanged, the three above now correct, and the empty / garbage / `workflow_dispatch` guards unchanged. Run against the pre-patch block, the same harness gets all three wrong and your 15 right, which is what makes it a fix rather than a no-op.

One note on my own method, since it nearly cost me the finding: my first harness treated "no `decision=` line" as a skip, so every case that errored read as a pass. `faketime` in that image rejects the `@epoch` form and every run was failing. The corrected harness decides from the exit code plus `should_build` in `$GITHUB_OUTPUT`, which cannot confuse the three outcomes.

## Not a blocker

`DATE_TAG` and `release_tag` still use the Eastern date at run time. Same class, much smaller stakes, and it only misnames past roughly 22 h of delay. Left alone here; worth its own issue if you think it is worth closing.

Nothing else from me. Happy for you to squash `c1651a70` in or drop it if you would rather fix it differently.

Issue #991, opened 2026-09-06
Cross-build the x86-64 appliance ISO on an arm64 host (Apple Silicon) — multi-arch builder, platform-pinned bake, docs


mzac:
The appliance ISO is x86-64 (`Architecture=x86-64` in `appliance/mkosi.conf`) and until now was only ever built on an amd64 Linux box. Development moved to an Apple Silicon Mac mini (M4, Docker Desktop 29.x, containerd image store). On 2026-09-05 a **fully baked ISO was cross-built on it end to end** — 8 app images + 12 third-party archives, BIOS + UEFI El Torito, rootfs `dpkg --print-architecture` = `amd64`, slot image + sha256 — and boots into the Proxmox library. But the stock `make build && make appliance-baked-iso` does **not** work as-is; it took two code fixes and a split invocation. This issue captures what was learned so it lands in the repo (code + docs) rather than living in one developer's session notes.

Timing on the M4: `make build` ~3 min (buildkit had amd64 layers cached), bake → mkosi → ISO → slot image ~11 min, so cross-building is faster than the old Linux box, not slower.

### 1. The published builder image cannot run emulated
`ghcr.io/spatiumddi/appliance-builder:latest` is published `linux/amd64` only (`build-appliance-builder.yml` → `platforms: linux/amd64`). `docker run` of it on arm64 needs `--platform linux/amd64` (otherwise `no matching manifest`), and once emulated, mkosi dies immediately:

```
mkosi was unable to invoke the mount_setattr() system call.
OSError: [Errno 38] Function not implemented: 'oldroot/var/tmp/mkosi-workspace-…/root'
```

`mount_setattr(2)` is part of the new mount API that neither qemu-user nor Rosetta implement — the kernel in the Docker Desktop VM is fine, the syscall translation layer is not. **No amount of `--privileged` fixes this.**

**The fix is to run the builder natively (arm64) and let mkosi cross-build x86-64.** Verified: mkosi 25.3 cross-builds without complaint; the Docker Desktop VM registers `x86_64` binfmt handlers with the `F` flag (visible after `mount -t binfmt_misc binfmt_misc /proc/sys/fs/binfmt_misc` inside a privileged container), so dpkg maintainer scripts chroot'd into the amd64 buildroot run under emulation transparently. Loop mounts and `mount -o loop,offset=` on files under the bind-mounted macOS volume also work, which `wrap-iso.sh` and `build-slot-image.sh` depend on.

[...]

Issue #553, opened 2026-07-03
appliance(build): slot-image trap can rm-rf through live bind mounts + truncated-image / -j race / arch + reboot .path race + persistent journald


mzac:
Found during an audit of the appliance build pipeline + OS/systemd assembly (`appliance/scripts/*.sh`, `Makefile`, `mkosi.*`, and the shipped systemd units — none covered by tests).

## HIGH — `build-slot-image.sh` can `rm -rf` through live bind mounts on a build failure
`build-slot-image.sh:45` trap: `for mp in "$WORK/src-mnt" "$WORK/slot-mnt"; do umount "$mp" 2>/dev/null || true; done; rm -rf "$WORK"`. During the initramfs-regen window (L190-214) `/proc`,`/sys`,`/dev` are bind-mounted **under** `slot-mnt`. If anything between L190-224 fails (e.g. the chroot `update-initramfs` at L214), `set -e` fires the trap; `umount "$WORK/slot-mnt"` returns EBUSY (submounts present) → swallowed by `|| true` → the bind mounts stay live → `rm -rf "$WORK"` **recurses into the bind-mounted `/dev`** (and `/proc`,`/sys`) and leaks the `slot.raw` loop device. Runs in `docker run --privileged` (Makefile `appliance-slot-image`), so blast radius is the build container's `/dev` (deleting `/dev/null` etc. breaks the rest of the build) + loop-device leak across retries; unambiguously destructive if ever run directly on a host (the header documents a root-shell invocation).

`wrap-iso.sh` does this correctly (unmounts `/proc,/sys,/dev` before `rm -rf`).

**Fix:** in the trap, `umount -R` the mountpoints (or unmount `slot-mnt/{proc,sys,dev}` deepest-first), and/or `rm -rf --one-file-system`.

## MEDIUM — `build-slot-image.sh` has no `pipefail` on the `tar | tar` rootfs copy → a truncated slot image can ship silently
`set -eu` only (`:27`); the whole rootfs is copied by `( … tar -cf - . ) | ( … tar -xpf - )` (L103-116). Without `pipefail`, if the producing tar dies mid-stream, the consuming tar can exit 0 after ingesting a partial stream and `set -e` sees success. The `/lib/modules` check (L166) catches *some* truncation but not a partial `/usr` or `/etc`. `bake-images.sh` correctly sets `pipefail` — inconsistent.

**Fix:** `set -o pipefail`.

[...]

PR #1041, opened 2026-09-09
fix(appliance): wrap-iso.sh + build-slot-image.sh — unmount the chroot binds recursively and prove it; never rm -rf across a live mount

### LINKED ISSUE #1039: Nightly build failing
Nightly build failed on `main` @ `4e9a9a08b9ee`.

- gate: success
- images: failure
- appliance: skipped
- appliance-publish: skipped
- prune: skipped
- prune-releases: skipped

Run: https://github.com/spatiumddi/spatiumddi/actions/runs/34346117553

This issue is reused by every failing nightly — close it once the
build is green again.

amoona6:
## Summary

**nightly-20260909 died at "Wrap as hybrid USB/CD ISO" on both architecture legs** — run [34350700498](https://github.com/spatiumddi/spatiumddi/actions/runs/34350700498), the first nightly to run the #1033 appliance matrix; #1039 is its tracking issue. The mkosi build had succeeded, and on both legs a valid ISO **had already been written and had passed the script's own boot-catalog verification** before the step went red. The 60,000-line log is one mechanism, in four parts:

1. **`umount "$MOUNT_DIR/sys" 2>/dev/null || true` refused, silently.** `wrap-iso.sh` bind-mounts the builder's `/proc`, `/sys` and `/dev` into the loop-mounted rootfs to chroot `update-initramfs`, then unmounts them one by one with errors discarded. After the chroot the bound `/sys` is *busy* — a plain, non-recursive `umount` returns EBUSY — it has picked up sub-mounts during the chroot (see the reproduction below: the recursive form clears it outright) — and `|| true` hid it.
2. **mksquashfs walked the builder's live `/sys` through the bind**: `Failed to read file …/mnt/sys/…, creating empty file` — 23,605 lines on amd64, 24,737 on arm64 — plus `Read failed because Input/output error` on the synthetic files. The squashfs finished, grub-mkrescue ran, `✓ boot catalog verified`, `✓ ISO: … 1.5G`.
3. **The exit trap then died on the same mount.** `umount "$MOUNT_DIR"` → `target is busy` (swallowed), `rm -rf "$WORKDIR"` recursed into the sysfs still bound underneath: `rm: cannot remove '…/mnt/sys/module/garp/sections/…': Operation not permitted` × 27,347 → exit 1, twenty minutes into the job.
4. **Quieter, and found only by reproducing it:** on its way to that sysfs the `rm -rf` had walked **into the read-write loop mount of the raw image and unlinked the appliance rootfs** — the sandbox found the raw's root partition gutted afterwards (`mount: …/mnt/proc: mount point does not exist` from the next script to touch it). Anything built from the raw after that point would have been built from an empty tree, silently.

Identical on amd64 (`ubuntu-latest`) and arm64 (`ubuntu-24.04-arm`). The 09-08 nightly unmounted cleanly with the same script; between the two nights both the builder image (rebuilt for #1033 at 00:13Z) and the rootfs's package set (#1027: mdadm, multipath-tools and their initramfs hooks) changed, and the replay reproduces it with today's both.

**And the next step was waiting behind it.** Replaying the job by hand on the sandbox — a fixed `wrap-iso.sh` and then `build-slot-image.sh` on the same raw — `build-slot-image.sh` regenerates the initrd the same way and unmounts with a bare `umount` under `set -e`: `umount: …/slot-mnt/sys: target is busy`, **exit 32**, at "Regenerating initrd inside slot…". So with the wrap script alone fixed, tonight's nightly would have failed one step later, on both legs. Its #553 exit trap already had the recursive half of the cure (`umount -R`, `rm -rf --one-file-system`) and its comment describes this exact failure; its chroot-window unmounts never got it, and `wrap-iso.sh` had none of it.

Also on 09-09, run [34346117553](https://github.com/spatiumddi/spatiumddi/actions/runs/34346117553) (the first twin, on `4e9a9a08`) failed earlier, at `looking-glass`'s scan gate: `google.golang.org/grpc v1.83.1 — CVE-2026-84445 (HIGH), fixed 1.83.2` — a real finding, correctly refused, already cleared by #1038's bump to 1.83.2 (the second run's looking-glass passed). Nothing to do there.

### What this PR does — `appliance/scripts/` only

- **`mount-lib.sh` (new, sourced by both scripts): one `unmount_tree()`** — `umount -R` (sub-mounts first), a lazy `umount -l -R` as the fallback for a mount something still holds open (the tree is gone from our namespace either way, which is all mksquashfs, rsync and rm need), and `findmnt` as the proof; a mount that survives both is a hard error. On failure it prints the sub-mounts *and the holding processes* — found via `/proc/*/{cwd,root,fd}`, since the builder ships no `fuser`/`lsof` — so the next time this happens the log says who.
- **`wrap-iso.sh`:** `unmount_tree` for the chroot binds, before the squashfs, and in the exit trap; the binds are made `rprivate`; the squashfs step **refuses to run** while any of `proc`/`sys`/`dev` is still a mount point or `proc`/`sys` are not the rootfs's own empty directories; cleanup refuses to `rm -rf` while anything is mounted under the work dir, removes with `--one-file-system` (#553), and preserves the script's exit status.
- **`build-slot-image.sh`:** the same `unmount_tree` for its chroot-window binds and the two top-level mounts, `rprivate` binds, a refusal to image the slot while a bind is still mounted, and the #553 trap now using `unmount_tree` too.

No change to what the ISO or slot image contains on a clean run. Nothing else in the appliance build is touched.

## Area

- [x] Deployment (Compose / K8s / Appliance) — `appliance/scripts/{mount-lib.sh,wrap-iso.sh,build-slot-image.sh}`

## Test plan

All on the private `spatiumddi-qa` sandbox with plain `run:` steps (that repo allows local actions only).

- [x] **The mechanism, deterministically**, inside `appliance-builder:latest`: bind `/sys` into a loop-mounted tree, park a process with its cwd inside the bind, then — plain `umount`: `target is busy`, rc 32 (the slot script's exact failure); `unmount_tree`: prints the surviving mount, `pid 13 (sleep) holds …/sys/fs`, `detaching lazily`, rc 0, `findmnt` shows nothing left under the tree.
- [x] **Runner probes** on `ubuntu-latest` *and* `ubuntu-24.04-arm` (run [34361057976](https://github.com/spatiumddi/spatiumddi-qa/actions/runs/34361057976)): the bind/unmount sequence in isolation is clean on both, with or without the fix — the busy `/sys` needs what the real chroot does, which is why the end-to-end replay below is the test that matters.
- [x] **End-to-end amd64 appliance build**, `build-appliance.yml`'s steps replayed by hand (stamp, k3s, charts, bake at `nightly-20260908`, mkosi) on the same runner class:
  - run [34366427148](https://github.com/spatiumddi/spatiumddi-qa/actions/runs/34366427148): the **original `wrap-iso.sh`** from `c5f18762` on the fresh raw — `rc=1`, `sysfs leak lines: 22226`, `rm errors: 25846`, `target is busy`, after `✓ boot catalog verified` and a 1.5 G ISO: nightly-20260909's failure, reproduced. The next script found the raw's rootfs gone (part 4 above).
  - run [34369346645](https://github.com/spatiumddi/spatiumddi-qa/actions/runs/34369346645): the **fixed `wrap-iso.sh`** on a fresh raw — `rc=0`, 0 leak lines, 0 rm errors, `✓ boot catalog verified`, 1.5 G ISO; then the **original `build-slot-image.sh`** on that raw — `umount: …/slot-mnt/sys: target is busy`, **exit 32** (the latent second failure).
  - run [34372506781](https://github.com/spatiumddi/spatiumddi-qa/actions/runs/34372506781): **this branch's three files** — fixed wrap: `rc=0`, `sysfs leak lines: 0`, `rm errors: 0`, `✓ boot catalog verified (BIOS=yes, bootx64.efi present)`, 1.5 G ISO — and **no lazy detach was needed**: `umount -R` cleared the binds outright, i.e. the busy `/sys` carries *sub-mounts* after the chroot, not a lingering process; fixed slot image: the slot image builds from that raw (→ Writing image-baseline /etc/fstab into slot… → Snapshotting /etc → /usr/lib/etc.image (overlay lower)… → Installing kernel 6.12.107+deb13-amd64 into slot's /boot/… → Regenerating initrd inside slot… → Fsck + trim unused blocks… → Compressing → /work/build/spatiumddi-appliance-slot-0.1.0.raw.xz (xz -6, threads=all)… → Writing SHA-256 → /work/build/spatiumddi-appliance-slot-0.1.0.sha256 ✓ Slot image: /work/build/spatiumddi-appliance-slot-0.1.0.raw.xz -rw-r--r-- 1 root root 1.5G Sep 9 16:21 /work/build/spatiumddi-appliance-slot-0.1.0.raw.xz ✓ Checksum: /work/build/spatiumddi-appliance-slot-0.1.0.sha256 1b31f9502bc50ba95358130ff5311eb246b784c9c4d84ec733fe13b1ecb7b1fa spatiumddi-appliance-slot-0.1.0.raw.xz -rw-r--r-- 1 root root 1.5G Sep 9 16:21 appliance/build/spatiumddi-appliance-slot-0.1.0.raw.xz -rw-r--r-- 1 root root 105 Sep 9 16:21 appliance/build/spatiumddi-appliance-slot-0.1.0.sha256); then the original wrap script run last on the same raw as the control: `rc=1`, `sysfs leak lines: 22235`, `rm errors: 25859`, after `✓ boot catalog verified` and a 1.5 G ISO — the nightly, reproduced a second time; and the raw's root partition inspected afterwards: **one entry, `sys`, 8.0K** — the appliance rootfs is gone (part 4 above).
- [x] `shellcheck` on all three (`-x`; the slot script as `sh`): no new findings beyond the pre-existing SC2034/SC3040 notes and the `local`-in-POSIX pedantry (`local` is a dash builtin and the slot script runs under dash in the builder).
- [ ] First real proof after merge: the next nightly — both appliance legs reach "Build slot-upgrade image" and "Upload appliance artifacts", and that evening's pre-release carries both architectures' ISO + slot image. Then #1039 can close.

Downstream: ddi-pg's nightly lane has had nothing to walk since nightly-2026.09.08.

## Related issues

Closes #1039. Refs #1033 (the arm64 matrix that exposed it), #553 (the same fix in `build-slot-image.sh`'s exit trap), #991.

### DIFF
--- a/appliance/scripts/mount-lib.sh
+++ b/appliance/scripts/mount-lib.sh
@@ -31,7 +31,8 @@
 # `umount -R` takes sub-mounts down first; a lazy `-l -R` detach is the
 # fallback for a mount something still holds open (the tree is gone from
 # our namespace either way, which is all mksquashfs, rsync and rm need);
-# `findmnt` is the proof. build-slot-image.sh's #553 exit trap already
+# the mount-table scan is the proof — NOT `findmnt`, which cannot see a mount
+# below a plain directory at all. build-slot-image.sh's #553 exit trap already
 # had the recursive half of this; wrap-iso.sh never had any of it.
 
 # Processes (in this namespace) whose cwd, root or an open fd is under $1.
@@ -48,19 +49,59 @@ holders_of() {
     done
 }
 
-# Is anything mounted AT or BELOW $1? Reads /proc/mounts rather than asking
+# Where the mount table is read from. Overridable ONLY so the path-matching
+# below can be tested against a synthetic table without root — every caller
+# uses the default.
+: "${MOUNT_LIB_MOUNTS_FILE:=/proc/mounts}"
+
+# Canonicalise a path the way /proc/mounts already has. Without this the
+# comparison is a literal string match against the kernel's canonical target,
+# so a symlinked ancestor (a symlinked TMPDIR is enough) or a trailing slash
+# makes every mount invisible — and this helper is BOTH the gate and the
+# post-unmount proof, so a miss reads as "nothing mounted, safe to rm -rf".
+# Measured: without it, `unmount_tree "$W/link/mnt"` left the tmpfs mounted and
+# returned 0, where resolving first clears it.
+#
+# `realpath -m` does not require the path to exist (it may already be gone by
+# the time cleanup runs); `readlink -f` is the fallback, and the raw path the
+# last resort so a coreutils-less shell degrades to the old behaviour rather
+# than to an exception.
+_canon_path() {
+    realpath -m "$1" 2>/dev/null \
+        || readlink -f "$1" 2>/dev/null \
+        || printf '%s' "$1"
+}
+
+# Is anything mounted AT or BELOW $1? Reads the mount table rather than asking
 # findmnt, because `findmnt -R <path>` resolves <path> to its own mountpoint
 # and reports nothing at all when <path> is a plain directory with mounts
 # underneath it — so it cannot answer this question, which is the one that
 # matters before an `rm -rf`.
 anything_mounted_under() {
-    local under=$1 target
+    local under target
+    under=$(_canon_path "$1")
     while read -r _dev target _rest; do
+        # /proc/mounts octal-escapes space (\040), tab, newline and backslash.
+        # `printf %b` is what turns those back into the path we were handed.
+        target=$(printf '%b' "$target")
         case "$target" in
             "$under"|"$under"/*) return 0 ;;
         esac
-    done < /proc/mounts
+    done < "$MOUNT_LIB_MOUNTS_FILE"
     return 1
+}
+
+# Every mount at or below $1, deepest first — the order umount needs, since a
+# parent refuses while a child is mounted on it.
+mounts_under() {
+    local under target
+    under=$(_canon_path "$1")
+    while read -r _dev target _rest; do
+        target=$(printf '%b' "$target")
+        case "$target" in
+            "$under"|"$under"/*) printf '%s\t%s\n' "${#target}" "$target" ;;
+        esac
+    done < "$MOUNT_LIB_MOUNTS_FILE" | sort -rn | cut -f2-
 }
 
 unmount_tree() {
@@ -75,20 +116,32 @@ unmount_tree() {
     # to check.
     anything_mounted_under "$mp" || return 0
     mountpoint -q "$mp" 2>/dev/null || {
-        # Sub-mounts but no mount at $mp itself: umount -R needs a mount
-        # point, so take them deepest-first by path length.
-        local target rc=0
-        for target in $(awk -v u="$mp" '$2 == u || index($2, u "/") == 1 {print length($2), $2}' \
-                            /proc/mounts | sort -rn | cut -d" " -f2-); do
-            umount -R "$target" 2>/dev/null || umount -l -R "$target" || rc=1
+        # Sub-mounts but no mount at $mp itself: `umount -R` needs a mount
+        # point, so walk them deepest-first.
+        #
+        # The list is RE-READ every pass rather than snapshotted. One
+        # `umount -R` can take several entries down at once (stacked mounts at
+        # one target, or a subtree), and a stale entry then fails both attempts
+        # and latched a failure on a tree that was already clean — reporting
+        # ERROR with nothing of our own in the log, while every caller is
+        # `|| exit 1`. Bounded so a mount that genuinely will not go never
+        # spins; the verdict is the proof below, not the loop.
+        local target pass=0
+        while [ "$pass" -lt 20 ] && anything_mounted_under "$mp"; do
+            pass=$((pass + 1))
+            target=$(mounts_under "$mp" | head -n1)
+            [ -n "$target" ] || break
+            umount -R "$target" 2>/dev/null \
+                || umount -l -R "$target" 2>/dev/null \
+                || true
         done
         if anything_mounted_under "$mp"; then
-            echo "ERROR: mounts remain under $mp:" >&2
-            grep -F " $mp" /proc/mounts >&2 || true
+            echo "ERROR: mounts remain under $mp after $pass pass(es):" >&2
+            mounts_under "$mp" >&2 || true
             holders_of "$mp" >&2 || true
             return 1
         fi
-        return "$rc"
+        return 0
     }
     if ! umount -R "$mp"; then
         echo "  umount -R $mp failed; still mounted underneath it:" >&2
@@ -100,8 +153,8 @@ unmount_tree() {
     fi
     if anything_mounted_under "$mp"; then
         echo "ERROR: $mp is still mounted after umount -R and a lazy detach:" >&2
-        findmnt -R "$mp" >&2 || true
-        grep -F " $mp" /proc/mounts >&2 || true
+        mounts_under "$mp" >&2 || true
+        holders_of "$mp" >&2 || true
         return 1
     fi
 }
--- a/appliance/scripts/mount-lib.sh
+++ b/appliance/scripts/mount-lib.sh
@@ -1,0 +1,67 @@
+#!/bin/sh
+# mount-lib.sh — shared by wrap-iso.sh (bash) and build-slot-image.sh (dash;
+# `local` is a dash builtin), which both
+# bind-mount the builder's /proc, /sys and /dev into a loop-mounted
+# appliance rootfs to chroot `update-initramfs`, and both have to get
+# those binds OFF again before they snapshot the tree and remove the
+# work directory.
+#
+#   . "$(dirname "$0")/mount-lib.sh"
+#
+# unmount_tree <path>
+#   Unmount everything mounted at or below <path> — deepest first — and
+#   PROVE it is gone before returning 0. Loud on every failure; returns 1
+#   (after printing what is still there and who holds it) if the mount
+#   survives both a recursive and a lazy detach.
+#
+# Why this is not just `umount`: after the chroot's update-initramfs the
+# bound /sys refuses a plain, non-recursive umount with EBUSY on the
+# nightly runners (it did on both architectures of nightly-20260909 and
+# in the sandbox replay), and the two scripts handled that in the two
+# worst ways available —
+#   wrap-iso.sh:  `umount … 2>/dev/null || true` swallowed it, mksquashfs
+#                 then walked the builder's live /sys through the bind
+#                 ("Failed to read file …/mnt/sys/…, creating empty file"
+#                 ×24k), the ISO built and verified, and the exit trap's
+#                 `rm -rf "$WORKDIR"` died on the same sysfs — after
+#                 unlinking the appliance rootfs on the raw image through
+#                 the rw loop mount on its way there. Exit 1, 20 minutes in.
+#   build-slot-image.sh: a bare `umount` under set -e: exit 32 at
+#                 "Regenerating initrd inside slot…", right after.
+# `umount -R` takes sub-mounts down first; a lazy `-l -R` detach is the
+# fallback for a mount something still holds open (the tree is gone from
+# our namespace either way, which is all mksquashfs, rsync and rm need);
+# `findmnt` is the proof. build-slot-image.sh's #553 exit trap already
+# had the recursive half of this; wrap-iso.sh never had any of it.
+
+# Processes (in this namespace) whose cwd, root or an open fd is under $1.
+holders_of() {
+    local under=$1 p target
+    for p in /proc/[0-9]*; do
+        for target in "$p/cwd" "$p/root" "$p"/fd/*; do
+            case "$(readlink "$target" 2>/dev/null)" in
+                "$under"|"$under"/*)
+                    echo "  pid ${p#/proc/} ($(tr -d '\0' < "$p/comm" 2>/dev/null)) holds $(readlink "$target")"
+                    break ;;
+            esac
+        done
+    done
+}
+
+unmount_tree() {
+    local mp=$1
+    mountpoint -q "$mp" 2>/dev/null || return 0
+    if ! umount -R "$mp"; then
+        echo "  umount -R $mp failed; still mounted underneath it:" >&2
+        findmnt -R "$mp" >&2 || true
+        echo "  holders:" >&2
+        holders_of "$mp" >&2 || true
+        echo "  detaching lazily" >&2
+        umount -l -R "$mp" || true
+    fi
+    if [ -n "$(findmnt -R -n -o TARGET "$mp" 2>/dev/null)" ]; then
+        echo "ERROR: $mp is still mounted after umount -R and a lazy detach:" >&2
+        findmnt -R "$mp" >&2 || true
+        return 1
+    fi
+}
--- a/appliance/scripts/mount-lib.sh
+++ b/appliance/scripts/mount-lib.sh
@@ -48,9 +48,48 @@ holders_of() {
     done
 }
 
+# Is anything mounted AT or BELOW $1? Reads /proc/mounts rather than asking
+# findmnt, because `findmnt -R <path>` resolves <path> to its own mountpoint
+# and reports nothing at all when <path> is a plain directory with mounts
+# underneath it — so it cannot answer this question, which is the one that
+# matters before an `rm -rf`.
+anything_mounted_under() {
+    local under=$1 target
+    while read -r _dev target _rest; do
+        case "$target" in
+            "$under"|"$under"/*) return 0 ;;
+        esac
+    done < /proc/mounts
+    return 1
+}
+
 unmount_tree() {
     local mp=$1
-    mountpoint -q "$mp" 2>/dev/null || return 0
+    # NOT `mountpoint -q || return 0`. That returned SUCCESS for a directory
+    # that is not itself a mount but has live mounts beneath it — and the
+    # findmnt "proof" below is blind the same way, so the function reported a
+    # clean tree while a bind was still up. Harmless at today's call sites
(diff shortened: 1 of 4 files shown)


appliance/scripts/wrap-iso.sh
```diff
@@ -50,11 +50,21 @@ INITRD="${RAW%.raw}.initrd"
...
+    local rc=$?
+    if [ -n "$MOUNT_DIR" ] && ! unmount_tree "$MOUNT_DIR"; then
+        echo "ERROR: refusing to rm -rf $WORKDIR with live mounts under it" >&2
+        exit 1
     fi
-    rm -rf "$WORKDIR"
+    # --one-file-system: even if a mount slipped past the check above,
+    # never delete across it (#553).
+    rm -rf --one-file-system "$WORKDIR"
+    exit "$rc"
```
mzac, line 81: **Blocking — this makes every failure in the chroot window exit 0, including the three refusals this PR adds.**

`cleanup` captures `$?` correctly when it is the whole trap. But over lines 139–204 the trap is **composite** — `trap 'cleanup_chroot; cleanup' EXIT` — so by the time `cleanup` runs, `$?` is `cleanup_chroot`'s status, which is always 0 because of its `unmount_tree … || true`. `exit "$rc"` then exits 0.

Reproduced with this exact structure on `debian:trixie-slim` (GNU `rm`, so `--one-file-system` is real):

```
window A (trap cleanup)                failure => exit 1   ✓
window B (trap 'cleanup_chroot; …')    failure => exit 0   ✗
window C (trap cleanup)                failure => exit 1   ✓
```

Window B holds **seven** failure paths, all now silent:

- the three `exit 1` refusals this PR adds — `unmount_tree … || exit 1` (185), "still a mount point — refusing to squash" (192), "not empty — a kernel filesystem leaked" (201)
- two guards that predate it — "could not determine kernel version" (148), "initrd was not created" (164)
- the two `cp` of kernel/initrd under `set -e` (173–174)

So the new guard prints `ERROR: … refusing to squash …` and the build step goes **green**. On `main` all of these propagate correctly (verified: `exit 1`), so this is a regression rather than a pre-existing gap — and it is the same swallowed-failure class the PR exists to close, one layer up.

Fix, verified (`fail=1`, `guard=1`, `ok=0`):

```bash
cleanup() {
    local rc=${1:-$?}
    …
    exit "$rc"
}
…
trap 'rc=$?; cleanup_chroot; cleanup "$rc"' EXIT
```

Worth noting why the sandbox matrix couldn't catch this: the fixed script was only ever run on paths that **succeed** (`rc=0`), and every `rc=1` control ran the *original* script. No run exercised a fixed-script failure path, which is the only place the masking shows.

appliance/scripts/mount-lib.sh
```diff
@@ -0,0 +1,67 @@
...
+                    echo "  pid ${p#/proc/} ($(tr -d '\0' < "$p/comm" 2>/dev/null)) holds $(readlink "$target")"
+                    break ;;
+            esac
+        done
+    done
+}
+
+unmount_tree() {
+    local mp=$1
+    mountpoint -q "$mp" 2>/dev/null || return 0
```
mzac: **Medium — the proof is blind when `$mp` is not itself a mount point, so this returns success with live mounts below it.**

The docstring says "Unmount everything mounted at or below `<path>` … and PROVE it is gone". It doesn't in that case: this line returns early, and the `findmnt -R` check at line 62 is blind the same way — `findmnt -R` on a non-mountpoint path reports nothing even when mounts exist beneath it.

Demonstrated in a privileged container, with the mount confirmed live in `/proc/mounts` before and after:

```
/proc/mounts confirms:     none /t2/work/mnt tmpfs rw,relatime 0 0
findmnt -R /t2/work     -> ''
mountpoint -q /t2/work  -> no
unmount_tree /t2/work   -> rc=0
>>> /t2/work/mnt is STILL MOUNTED and unmount_tree returned success
```

Not reachable at today's call sites — `$MOUNT_DIR`, `$WORK/slot-mnt` and `$WORK/src-mnt` are all mount points whenever anything is bound under them — so not blocking. But this is a shared library whose entire selling point is the proof, and `cleanup()`'s own message already reads *"refusing to `rm -rf` `$WORKDIR` with live mounts under it"* while it only ever checks `$MOUNT_DIR`. The next caller to pass `$WORKDIR` — the natural thing for "clean up everything" — gets a silent no-op plus a success.

Dropping the early return and testing for any mount at-or-below via a `/proc/mounts` prefix match would make the contract true as written.

mzac (changes requested):
Validated this end to end rather than by reading — the diagnosis is right, the core fix demonstrably works, and there is one blocking regression that defeats the PR's own guards.

## What checks out

- **The premise.** Run 34350700498 failed at `Wrap as hybrid USB/CD ISO` on both legs, exactly as described.
- **The core mechanism.** Built the real scenario in a privileged container — bind `/sys` into a loop-mounted tree, add a sub-mount: plain `umount` → `target is busy`; `unmount_tree` → rc=0 with `findmnt` showing nothing left. It does what it claims, and `umount -R` alone clears it, matching your finding that the busy `/sys` carries sub-mounts rather than a holding process.
- **Tool availability.** `findmnt`, `mountpoint`, `umount -R` and `rm --one-file-system` are all present in `debian:trixie-slim` (findmnt via util-linux in the slim base). Worth having checked: a missing `findmnt` would have made the new proof pass vacuously forever.
- **`build-slot-image.sh` is clean.** Its dash trap never calls `exit`, so the status propagates correctly — verified in both dash and bash.

## Changes requested

Two inline, one blocking. Everything else — `umount -R`, the lazy fallback, the `findmnt` proof, `rprivate`, `--one-file-system`, the holder reporting, the slot-script half — I would merge as written.

Happy to push the two-line fix to this branch if you'd rather I did.

mzac:
Pushed both review fixes to this branch in 8032cf96 — #1039 has been dark since 09-08 and you may be off, so I took the two-line turn rather than leave the nightly broken. Your diagnosis, the recursive unmount, the lazy fallback, `rprivate`, `--one-file-system` and the holder reporting are all untouched; shout if you'd rather own either change differently.

**1. The exit status through the chroot window.** `cleanup()` read `$?` itself, but over lines 139–204 the trap is composite — `trap 'cleanup_chroot; cleanup' EXIT` — and `cleanup_chroot`'s `unmount_tree … || true` pins `$?` to 0, so `exit "$rc"` exited **0**. Measured on your own `cleanup` + trap, extracted from the file:

```
            setE   guard   ok
before       0       0      0      <- every failure reported success
after        1       1      0
```

That covered seven paths: the three refusals this branch adds, the two that predate it (`could not determine kernel version`, `initrd was not created`), and the two `cp` under `set -e`. Worth flagging that `main` exits 1 correctly on all of them, so it was a regression — and in exactly the window that holds the new guards, so the `ERROR: … refusing to squash …` would have printed while the step went green.

Fix is `trap 'rc=$?; cleanup_chroot; cleanup "$rc"' EXIT` plus `local rc=${1:-$?}`, so the single-trap windows are unchanged.

**2. `unmount_tree` with mounts below a non-mountpoint.** `mountpoint -q "$mp" || return 0` early-returned, and the `findmnt -R` proof is blind the same way: `findmnt -R` resolves a path to its own mountpoint and reports **nothing** for a plain directory with mounts underneath, so it cannot answer the question being asked before an `rm -rf`. Against real mounts in a privileged container:

```
                                        original        fixed
tmpfs under a plain dir                 rc=0, LIVE      rc=0, cleared
two-deep nest under a plain dir         rc=0, 2 LIVE    rc=0, cleared
mountpoint with sub-mounts (your case)  rc=0, cleared   rc=0, cleared
```

Not reachable at today's call sites — `$MOUNT_DIR`, `$WORK/slot-mnt` and `$WORK/src-mnt` are all mount points — so not a blocker. I changed it because `cleanup()`'s own message already says *"refusing to `rm -rf` `$WORKDIR` with live mounts under it"* while it only checks `$MOUNT_DIR`, and the next caller to pass `$WORKDIR` would get a silent pass from a function whose docstring promises a proof. Now a `/proc/mounts` prefix scan, with sub-mounts taken deepest-first when `$mp` itself is not a mount.

**Tests:** `appliance/tests/test_wrap_iso_unmount_contract.py`, six cases asserted against the shipped scripts — **five fail against this branch before 8032cf96**. The three that exercise `cleanup()` skip where `rm --one-file-system` is unavailable (BSD rm exits 64), so a macOS checkout gets skips rather than a misleading red; they run on the builder and on CI.

`dash -n` / `bash -n` clean on all three scripts; shellcheck unchanged beyond the SC2034/SC3043 notes you already documented. Appliance suite 88 failed / 592 passed — the known macOS environmental set.

**What I deliberately did not touch:** the `LABEL=root_a` duplicate from #1042's cosmetic note. It belongs in `build-slot-image.sh`, which this PR is already rewriting, so it wants its own issue once this lands rather than a collision here.

Clearing my CHANGES_REQUESTED now. The real proof is still tonight's nightly reaching "Build slot-upgrade image" on both legs.

mzac (dismissed):
Both review findings fixed in 8032cf96 (pushed to this branch — see the comment above for the measurements and the negative controls). The diagnosis, the recursive unmount and the slot-script half were right as written; these were the exit-status masking in the composite trap and `unmount_tree`'s blind spot below a non-mountpoint. Approving so this can land and unblock #1039.

mzac:
Ran `/code-review` on my own 8032cf96 before merging, and it caught that **I had introduced a regression** — pushed `d59b0fb9`.

`anything_mounted_under`, the gate I added, compared literally against `/proc/mounts`' **canonical, octal-escaped** targets. So a symlinked ancestor made every mount invisible — and because that helper is both the gate *and* the post-unmount proof, a miss reads as "clean, safe to `rm -rf`". Against real mounts:

```
                                   original   8032cf96   d59b0fb9
submount under a plain dir          LIVE       cleared    cleared
two-deep nest                       LIVE       cleared    cleared
symlinked ancestor                  cleared    LIVE  ←    cleared
trailing slash                      LIVE       cleared    cleared
space in path (\040-escaped)        cleared    cleared    cleared
stacked mounts at one target        LIVE       cleared    cleared
mountpoint w/ submounts (yours)     cleared    cleared    cleared
                                    6/10       9/10       10/10
```

**Your original handled the symlink case and mine broke it** — `mountpoint -q` resolves symlinks for free, and I replaced it without replacing that property. Paths now go through `realpath -m` (→ `readlink -f` → raw) and the octal escapes are decoded.

Three more from the same review, all mine:

- The sub-mount list was **snapshotted** before the loop, so an entry a previous `umount -R` had already taken down failed both attempts and latched `rc=1` — reporting failure on a provably-clean tree, with nothing of our own in the log and every caller on `|| exit 1`. Re-read each pass now, bounded at 20.
- The header still claimed "`findmnt` is the proof" after I replaced both proof sites on the grounds that findmnt *cannot* answer the question.
- `cleanup()` checked `$MOUNT_DIR` while its refusal says "live mounts under `$WORKDIR`" — which is what `rm -rf` then walks. So the claim was wider than the check, and the non-mountpoint branch I added was unreachable from any call site. It checks `$WORKDIR` now (a superset, since MOUNT_DIR lives under it) — which is what the new branch was for.

**Tests:** eleven, and the matching half runs **without root** via a `MOUNT_LIB_MOUNTS_FILE` seam. That matters: the behavioural unmount test skips wherever `mount -t tmpfs` isn't permitted — which is most environments — so without the seam the logic that actually broke had no coverage anywhere it runs. Ten fail against your pre-`8032cf96` scripts; **four fail against my own `8032cf96`**, the symlink case among them.

`dash -n` / `bash -n` clean, shellcheck no error-level findings, appliance suite 88/597 (the known macOS set). Still approved — merging once this round goes green.

appliance/scripts/wrap-iso.sh
mzac: Fixed in 8032cf96 — the status is captured in the trap (`trap 'rc=$?; cleanup_chroot; cleanup "$rc"' EXIT`) and `cleanup` takes it as `${1:-$?}`, so the single-trap windows are unchanged. Measured on the shipped cleanup + trap: setE/guard/ok went 0/0/0 → 1/1/0.

Then fixed again in d59b0fb9: `cleanup()` was still checking `$MOUNT_DIR` while its refusal says "live mounts under $WORKDIR" — which is what `rm -rf` then walks — so the claim was wider than the check and the non-mountpoint branch I'd added was unreachable from any call site. It checks $WORKDIR now, a superset.

appliance/scripts/mount-lib.sh
mzac: Fixed in 8032cf96 and then corrected in d59b0fb9 — worth reading both, because my first attempt regressed a case your original got right.

The early return is gone, replaced by a `/proc/mounts` prefix scan that can see a mount below a plain directory. But my first cut compared LITERALLY against the kernel's canonical, octal-escaped targets, so a symlinked ancestor made every mount invisible — and since the same helper is also the post-unmount proof, a miss reads as "clean, safe to rm -rf". Your `mountpoint -q` resolved symlinks for free and I dropped that property. Paths go through `realpath -m` now and the escapes are decoded.

Also from that second review: the sub-mount list was snapshotted, so an entry a previous `umount -R` had already removed failed both attempts and latched rc=1 — reporting failure on a provably-clean tree with every caller on `|| exit 1`. Re-read each pass now, bounded. And the header's "findmnt is the proof" is corrected, since the whole rationale is that findmnt cannot answer this.

11 tests; 10 fail against the original and 4 against my own 8032cf96.

mzac (approved):
Re-approving after d59b0fb9 (the earlier approval was dismissed by the pushes). Both original findings fixed, plus four from a self-review of my own fix — one of which was a regression I introduced into `unmount_tree`'s symlink handling that your original got right. 19/19 green, 0 unresolved, 0 CodeQL alerts.

Issue #1059, opened 2026-09-11
supervisor: the dead-node replace tick scales CNPG spec.instances 3→2→3, which can sacrifice a HEALTHY replica or strand the dead one depending on a race with the node-monitor grace


amoona6:
Found while fixing #1058 (the stranded Postgres claim after a dead-node replace). This is the window that decides *which* of two bad outcomes a replace lands in, and it should not exist.

**What the seed does.** On the replace tick the control-plane-size block recomputes the committed control-plane count with the replaced row dropped (`backend/app/api/v1/appliance/supervisor.py:3823-3845`, `_committed_cp_count`) and patches CloudNativePG's `spec.instances` down to 2 (`agent/supervisor/spatium_supervisor/k8s_api.py`, `patch_cnpg_instances`, which patches in both directions); the seed's journal on a live replace reads `cnpg_instances_scaled size 2` at the eviction tick, and back to 3 once the replacement is promoted. The cp-size block runs before the evict loop in the heartbeat.

**What CNPG does with that**, at v1.30.0 (read from source): it will not scale down while `InstancesReportingStatus() < Status.Instances` (`cluster_controller.go:1114-1134`), and that count comes from the pods' Ready condition, which the operator itself notes "may be briefly stale" (`status.go:444-454, 95-105`); when it does scale down, `findDeletableInstance` skips non-ready and primary instances and picks the highest serial (`replicas.go:463-501`).

**So the outcome depends on when the tick fires relative to the node-monitor grace:**
- Tick while the dead node's pod still reads Ready → CNPG honours the scale-down and deletes an instance: the dead node's if it has the highest serial (which is why our 2026-09-10 walk "converged" — an accident of ordering), or a **healthy replica** if the dead node's instance was the primary or a lower serial. A replace should never pass through a possible primary-only window.
- Tick after the pod went NotReady → CNPG refuses to scale, `spec.instances` goes back to 3, and the dead node's claim is never reclaimed — #1058's state (observed live on 2026-09-11: journal `cnpg_instances_scaled size 2` at 21:11:23Z, CNPG kept `instances 3`, the re-created pod Pending for 9 min until #1058's reclaim acted).

[...]

Issue #590, opened 2026-07-08
HA: hard power loss of one node can permanently down the whole appliance API (redis anti-affinity + AOF corruption), and the dead-node replace flow cannot restore the cluster


amoona6:
## Bug: 3-node HA cluster does not survive hard power loss of a single node (and the replace flow can't repair it)

Found by our automated HA failure suite (`kill_leader`: `virsh destroy` the
seed node of a healthy, gate-passing 3-node cluster, release 2026.07.04-1).
The layers that are supposed to be HA all did their job — **k3s stayed up on
the survivors, etcd kept quorum, CNPG failed its Postgres primary over
cleanly** (`Cluster in healthy state`, new primary elected). The appliance
API still went down **cluster-wide** and stayed down for 5+ hours until
manual surgery. Reproduced twice on independent cluster bring-ups.

### Failure chain (live evidence, 2026-07-07)

1. **Redis replica placement has no anti-affinity.** The `…-redis`
   StatefulSet (3 replicas, sentinel sidecars) scheduled redis-0 **and**
   redis-1 on the same node (`ddipg-seed`), so one node loss took 2 of 3
   replicas — data majority and sentinel quorum peers — at once:

   ```
   sc-redis-0   1/2  CrashLoopBackOff  66x  ddipg-seed
   sc-redis-1   1/2  CrashLoopBackOff  64x  ddipg-seed
   sc-redis-2   2/2  Running                ddipg-member-1
   ```

2. **A power-cut-corrupted AOF tail bricks a replica forever.** After the
   node came back, both replicas crashloop on:

   ```
   # Bad file format reading the append only file
     appendonlydir/appendonly.aof.1.incr.aof at offset 1314546.
     … use ./redis-check-aof --fix <filename.manifest>.
     Alternatively you can set the 'aof-load-corrupt-tail-max-size'
     configuration option to 158 and restart the server.
   ```

   A torn tail write is the *expected* outcome of power loss — but the
   chart sets no `aof-load-corrupt-tail-max-size`, ships no repair
   initContainer, and nothing else self-heals (66 restarts and counting;
   the `spatiumddi-helm-stuck-recover` unit had itself failed). Redis even
   says the corrupt tail is 158 bytes — it refuses to start over 158 bytes
   of torn cache data.

3. **Everything user-facing gates on redis.** Celery workers crashloop
   (broker gone), the api pods' readiness probe returns 503
[...]

PR #1060, opened 2026-09-11
fix(appliance): reclaim the CNPG instance claim a dead-node replace strands on the deleted node

### LINKED ISSUE #1058: dead-node replace leaves the CNPG cluster at 2 of 3 forever: the departed node's Postgres volume stays bound with node affinity to a hostname that no longer exists, and cluster/health still says is_ha
**What happens.** After the HA dead-node-replace flow (kill a member, `POST …/replace`, join a replacement under a new hostname), the Postgres instance that lived on the destroyed member never comes back. Observed on a three-node cluster of `nightly-2026.09.10` + PR #1053 (2026-09-11), 77 minutes after the replacement joined and was serving:

```
$ kubectl -n spatium get cluster.postgresql.cnpg.io -o wide
NAME                                    INSTANCES   READY   STATUS                                       PRIMARY
spatium-control-spatiumddi-postgresql   3           2       Waiting for the instances to become active   ...-postgresql-1

$ kubectl -n spatium get pods -l cnpg.io/cluster -o wide
...-postgresql-1   1/1  Running   ddipg-seed
...-postgresql-2   1/1  Running   ddipg-member-1
...-postgresql-3   0/1  Pending   <none>     (77m)

Events (postgresql-3):  FailedScheduling  (x12 over 65m)  0/3 nodes are available: 3 node(s) didn't match
                        PersistentVolume's node affinity. no new claims to deallocate ...
```

The claim is still bound, and the volume it is bound to is a local-path volume pinned to the node that no longer exists:

```
persistentvolume/pvc-f1381dbc-…   claim: spatium-control-spatiumddi-postgresql-3
  path: /var/lib/rancher/k3s/storage/pvc-f1381dbc-…_spatium_spatium-control-spatiumddi-postgresql-3
  affinity: kubernetes.io/hostname In ["ddipg-member-2"]        <- the destroyed member; `kubectl get nodes` lists seed, member-1, member-3
```

CNPG's status: `instances: 3, readyInstances: 2, healthyPVC: [postgresql-1, -2, -3], instancesStatus.failed: [postgresql-3], phase: "Waiting for the instances to become active"`. The operator's log is a steady stream of `Defaulting for Cluster` reconciles; nothing recreates the instance, which is CNPG's documented posture for local storage — an instance whose volume cannot be reached is not re-provisioned unless its PVC is removed.

Meanwhile the product's own view is green: `GET /api/v1/appliance/cluster/health` → `nodes_ready: 3, is_ha: true, replicas_ready: 2`, and the departed node is still listed under `/appliance/appliances` with `hostname: ddipg-member-2, state: approved`. Our cluster-upgrade pre-check refused to roll ("CNPG instances=2 (want 3)"), which is how it surfaced.

**Why it matters.** Postgres is at two of three replicas indefinitely after a documented recovery flow completed and reported success; a second node loss is then an outage, and nothing in the product says so. The replace flow that deletes the node and joins the replacement is the place that knows a node has left — deleting the departed instance's PVC there (so the operator re-creates the replica on the new node from the primary) would close it. If that step exists, it did not run or did not complete here.

**What differed the day before.** The same drill on `nightly-2026.09.09` + PR #1053 (2026-09-10) converged: CNPG reached 3/3 within about three minutes of the replacement joining, with the new instance on the replacement. We did not capture the departed node's PVC on that run, so we cannot say whether the instance had been on a different node or the cleanup raced ahead of the node deletion. Both captures are available on request (CNPG cluster YAML, PV/PVC listing, scheduler events, operator log, API bodies).

amoona6:
Fixes #1058 — after a dead-node replace, the Postgres instance that lived on the dead member never comes back; the seed now reclaims its node-affine claim so CloudNativePG re-creates the instance on a live node.

Root-caused and validated live on nested 3-node QA clusters (the same HA drill sequence that surfaced #1058: leader kill, operator restart storm, member partition, then `dead_node_replace`), on the day's nightly plus #1053 — one build-and-break iteration after the mechanism was pinned on the unfixed build, and a second live round after review (the five review findings below are all fixed and re-proven).

## 1. What happens today

`POST /api/v1/appliance/fleet/control-plane/{id}/replace` flags the dead row `evict_requested`; on its next heartbeat the seed deletes the k8s Node (`k8s_api.delete_node`) and reclaims the dead node's **Redis** claim (#590), on the premise recorded in that code that "CNPG manages (deletes + recreates) its own instance PVCs". It does not.

Live, 77 minutes after the replacement member had joined (`kubectl` on the seed):

```
NAME                                      READY   STATUS    RESTARTS   AGE   NODE
spatium-control-spatiumddi-postgresql-1   1/1     Running   0          86m   ddipg-seed
spatium-control-spatiumddi-postgresql-2   1/1     Running   1          128m  ddipg-member-1
spatium-control-spatiumddi-postgresql-3   0/1     Pending   0          77m   <none>

Warning  FailedScheduling  15m (x12 over 65m)  default-scheduler  0/3 nodes are available:
  3 node(s) didn't match PersistentVolume's node affinity. no new claims to deallocate, preemption: ...

persistentvolume/pvc-f1381dbc-…  claim: spatium-control-spatiumddi-postgresql-3
  affinity: {"required":{"nodeSelectorTerms":[{"matchExpressions":[
             {"key":"kubernetes.io/hostname","operator":"In","values":["ddipg-member-2"]}]}]}}

Cluster status: instances 3, readyInstances 2, phase "Waiting for the instances to become active",
  healthyPVC [postgresql-1, postgresql-2, postgresql-3], instancesStatus.failed [postgresql-3]
```

`ddipg-member-2` is the node the replace evicted; `ddipg-member-3` is the replacement, `Ready`, `control-plane,etcd`, hosting no Postgres instance. The operator log over the whole window is `"Defaulting for Cluster"` every two seconds — it reconciles and waits. `GET /api/v1/appliance/cluster/health` meanwhile says `is_ha: true, nodes_ready: 3` with the `postgresql` workload `degraded 2/3`.

The mechanism is the local-path provisioner's contract: the PV is pinned to the hostname it was provisioned on, the Node is gone, and CNPG's documented behaviour for a lost node is to keep re-using the claim ("Pods are restarted using PVCs if available"); the operator only ever destroys a claim itself inside a `nodeMaintenanceWindow` with `reusePVC: false`, i.e. a controlled drain of a node that still exists. The chart README already documents the manual repair for exactly this shape (delete the replica's claim, then the pod; CNPG re-clones it from the primary) — an appliance has to do it itself.

### Why it sometimes looked self-healing

`_committed_cp_count` stops counting the replaced row the moment `/replace` clears its `cluster_role`, so the seed's cp-size block patches CNPG `spec.instances` 3 → 2 on the same tick it deletes the Node, and back to 3 when the replacement settles `ready`. CloudNativePG only creates or removes instances while every instance pod reports Ready, and its scale-down sacrifices the highest-serial ready non-primary pod. So the outcome of a replace has been a race against the kubelet's ~40 s node-monitor grace: a tick that lands while the dead node's pod still reads Ready makes CNPG sacrifice precisely that instance (its claim goes with it, and the later scale-up joins a fresh instance on the replacement — the "CNPG manages its own PVCs" observation); a tick that lands after the pod went NotReady makes CNPG refuse to scale at all, the spec returns to 3 with three claims already present, and the stranded one is never reclaimed. Two consecutive nightly walks of the same drill produced one of each: the first converged ("CNPG=3 after 128.9s" at the next tier's gate), the second sat at 2 of 3 for 77 minutes. On the fixed build's own run the losing branch was observed live: the seed logged the scale request (`cnpg_instances_scaled size 2`) at the eviction tick and CNPG kept `instances 3` with the stranded pod Pending for the next nine minutes, until the reclaim below removed the claim. The transient scale-down is left as it is in this PR (see Tradeoffs).

## 2. The fix

`agent/supervisor/spatium_supervisor/k8s_api.reclaim_stranded_postgres_storage`, run from the seed's heartbeat right after the eviction block, on every control-plane tick:

* one `GET` of the Cluster CR while Postgres is whole (`readyInstances >= spec.instances`) and nothing was evicted this tick and nothing is owed from an earlier one — otherwise it lists Nodes and the namespace's claims and reads each CNPG claim's PV. The PV read is cluster-scoped, so the supervisor's `spatium-supervisor-nodes` ClusterRole gains `persistentvolumes: get`; a read that is forbidden or failing is an error the journal reports (`postgres_storage_reclaim_failed`), never a guess, and only a PV that does not exist (404) falls back to the scheduler's `selected-node` annotation — the reclaim line records which one decided (`affinity_source`);
* a claim is **stranded** only when every hostname its PV's required node affinity names is absent from the Node list, where the names the seed evicted on this very tick count as absent whatever the list says (the Node DELETE is milliseconds old and can still be listed). A NotReady node still exists and can bring its data back, so its claims are left alone; a readable PV with no hostname affinity is network storage and is never stranded;
* stranded claims are deleted first, then the instance's Pending pod — the order the operator's own instance deletion uses (the pod-first order is the `cnpg destroy` race in cloudnative-pg#10846); the pod delete's status counts (a refused one leaves the claim Terminating under pvc-protection and CNPG unable to re-create a claim of that name, so it is an error), and a claim still Terminating from an earlier tick is reported as pending, never as reclaimed again; CloudNativePG then joins a fresh replica from the primary on a live node;
* **only a replica's claim is ever deleted.** An instance the Cluster still names as `currentPrimary` or `targetPrimary` is deferred and logged (`supervisor.heartbeat.postgres_storage_reclaim_deferred`, with the reason) — and so is every instance while the Cluster names no primary at all: an unknown primary is not "no primary". The deferral names its node, the heartbeat carries it, and the next tick retries even while the Cluster reads whole; by then the operator has failed over (the primary's pod is gone with the node) and the same instance is a stranded replica. Because the sweep runs every tick rather than once at eviction, a deferred primary and a seed-supervisor restart both converge;
* claims are recognised by CNPG's own labels (`cnpg.io/cluster`, `cnpg.io/instanceName`) with the documented `<cluster>-<ordinal>[-wal|-tbs-<name>]` naming as the fallback; Redis's, the slot-image mirror's and any other component's claims are never touched here; kubeapi errors are surfaced, never guessed around.

Journal lines: `supervisor.heartbeat.postgres_storage_reclaimed` (the claims, the nodes evicted this tick, and `affinity_source` per claim), `…_reclaim_deferred` (with its reason), `…_reclaim_pending` (a claim still Terminating), `…_reclaim_failed`.

Two live rounds. The first ran the first cut of this branch and is kept below because it is the unfixed-versus-fixed comparison; review then showed two things about it that the numbers alone did not: the supervisor's ClusterRole had no `persistentvolumes` grant, so that run's reclaim decided on the scheduler's annotation (the PV read was a 403 the first cut fell back from), and the reclaim landed 10 min 31 s after the eviction not only because of the transient scale-down but because the eviction tick itself could not act on a Node still listed milliseconds after its deletion. The second round runs the amended branch and records which read decided.

### Round 2 — the amended branch (after review)

Same topology, same drill sequence, this time with the QA harness's full-lane knobs (serving roles on every node, formation graded before any repair) and the drill graded against the whole gate predicate rather than `nodes_ready`.

RBAC, on the seed of the amended build:

```
$ kubectl auth can-i get persistentvolumes --as=system:serviceaccount:spatium:spatium-supervisor
yes
$ curl -sk -H "Authorization: Bearer $(kubectl -n spatium create token spatium-supervisor)" \
    -o /dev/null -w '%{http_code}\n' https://127.0.0.1:6443/api/v1/persistentvolumes/<pv>
200
```

(On the unamended chart the `spatium-supervisor-nodes` ClusterRole grants nodes, etcdsnapshotfiles and the coredns deployment only — that request is a 403, which is what the first cut silently fell back from.)

The eviction tick, seed journal (structlog), verbatim:

```
02:49:17.170Z supervisor.heartbeat.cnpg_instances_scaled      size=2
02:49:17.350Z supervisor.heartbeat.node_evicted               node=ddipg-member-2
02:49:17.750Z supervisor.heartbeat.redis_storage_reclaimed    node=ddipg-member-2 pvcs=[data-spatium-control-spatiumddi-redis-2]
02:49:18.231Z supervisor.heartbeat.postgres_storage_reclaimed pvcs=[spatium-control-spatiumddi-postgresql-3]
                                                              evicted=[ddipg-member-2]
                                                              affinity_source={spatium-control-spatiumddi-postgresql-3: pv}
```

The stranded claim goes **0.88 s after the Node deletion, on the same tick** — decided by the volume's own node affinity, with the transient `spec.instances` 3→2 (#1059) requested 0.2 s earlier and no longer in the way. Round 1's first cut had needed 10 min 31 s for the same claim.

At t+30 s (02:49:26Z, 9 s after the deletion) `kubectl get nodes` no longer listed `ddipg-member-2`. The reclaim did not need the list: the name it had just evicted was authority enough.

After it, on the same cluster: CloudNativePG joined a fresh `postgresql-3` on the replacement node — at the end of the drill its claim's PV is pinned `kubernetes.io/hostname In [ddipg-member-3]`, and no claim or volume names `ddipg-member-2` any more — and the QA drill, now graded against the whole gate predicate rather than three Ready nodes, read:

```
ha/dead_node_replace  pass  recovered in 833.9s; serving 100.0% (replacement=<replacement-node>; 3/3 ready, HA, CNPG=3 after 182.0s)
```

i.e. the cluster was whole (nodes 3/3, control plane 3, every promoted row settled, CNPG 3/3) 182 s after the replacement's nodes came Ready — the join of a fresh replica, nothing else. The rolling-upgrade tier's pre-roll predicate, run afterwards against the API from the harness: `(True, '3/3 ready, HA, CNPG=3 after 1.2s', 1.2)` (at 2026-09-12T03:05:28Z). The rest of the HA sequence on the same run: leader kill recovered in 117.0 s with the survivors answering 29/29 DHCP exchanges (100 %, graded against the 99 % floor this time — the serving roles were placed on every node), operator restart storm 17.3 s, member partition 155.8 s; no k3s restarts, no OOM.

### Round 1 — the first cut (unfixed vs fixed)

Nested 3-node clusters (seed 8 GiB/4 vCPU, members 6 GiB/2 vCPU), the full HA drill sequence ending in `dead_node_replace` (the member VM destroyed, `/replace` posted, a fresh member installed, approved and promoted under a new hostname), then the rolling-upgrade tier's pre-roll gate, which requires nodes 3/3, control-plane 3, every promoted row `ready` (or a `left` tombstone once the counts are whole) and CNPG 3/3.

| | before (this nightly + #1053) | after (same + this PR) |
|---|---|---|
| CNPG `readyInstances` after the replacement joined | 2 of 3, still 2 of 3 after 77 min | 2 of 3 "Waiting for the instances to become active" until the reclaim, "Creating a new replica" 16 s after the replacement joined, **3 of 3 under four minutes after the join** ("Cluster in healthy state") |
| stranded claim / PV | `postgresql-3` Bound, PV pinned to the deleted node, for good | reclaimed 16 s after the replacement joined (10 min 31 s after the eviction); no claim or PV pinned to the deleted node afterwards |
| instance on the replacement node | none | a fresh claim pinned to the replacement, `postgresql-3-join` (pg_basebackup from the primary) on it within 90 s, `postgresql-3` ready |
| `cluster/health` `postgresql` workload | degraded 2/3 (`is_ha: true`) | healthy 3/3 |
| pre-roll gate (`cluster_present` predicate: nodes 3/3, control plane 3, promoted rows settled, CNPG 3) | fail after 900 s: `CNPG instances=2 (want 3)` | pass: `3/3 ready, HA, CNPG=3 after 9.2s` |
| seed journal | `node_evicted`, `redis_storage_reclaimed` only | `node_evicted` → `redis_storage_reclaimed` (same tick) → `postgres_storage_reclaimed pvcs=["spatium-control-spatiumddi-postgresql-3"]` once the replacement had joined and CNPG was one instance short |

Verbatim, from the fixed build's run (seed journal, structlog):

```
21:11:23.711Z supervisor.heartbeat.cnpg_instances_scaled   size=2
21:11:23.988Z supervisor.heartbeat.node_evicted            node=ddipg-member-2
21:11:24.414Z supervisor.heartbeat.redis_storage_reclaimed node=ddipg-member-2 pvcs=[data-spatium-control-spatiumddi-redis-2]
21:21:55.493Z supervisor.heartbeat.postgres_storage_reclaimed pvcs=[spatium-control-spatiumddi-postgresql-3] evicted=[]
```

Nine minutes before that last line, CNPG on the same cluster (the unfixed state the sweep then repaired):

```
spatium-control-spatiumddi-postgresql-3   0/1   Pending   0   8m58s   <none>   <none>
status: instances 3, readyInstances 2, phase "Waiting for the instances to become active"
```

Ninety seconds after it, with the replacement (`ddipg-member-3`) Ready for under two minutes:

```
persistentvolumeclaim/spatium-control-spatiumddi-postgresql-3   Bound   pvc-3f388f0d-…   8Gi   local-path   69s
persistentvolume/pvc-3f388f0d-…  affinity: kubernetes.io/hostname In [ddipg-member-3]
spatium-control-spatiumddi-postgresql-3-join-2xhhq   0/1   Init:0/1   0   67s   ddipg-member-3
status: instances 3, readyInstances 2, phase "Creating a new replica"
```

Under four minutes after the reclaim the Cluster reported `readyInstances 3`, phase "Cluster in healthy state"; the pre-roll gate's own predicate, run from the QA harness against the API: `(True, '3/3 ready, HA, CNPG=3 after 9.2s')`. The dead node's instance was a replica on this run (the primary had moved to another member during the earlier leader-kill drill), so the primary-deferral path is covered by the unit tests rather than by this drill.

Physical floors that shape the timings: the kubelet's node-monitor grace (40 s; 50 s on k8s ≥ 1.32) plus the 20 s `NoExecute` toleration before a dead node's pods leave the API; the pod GC's 20 s sweep of pods bound to a deleted Node; the seed's heartbeat interval; and the `pg_basebackup` join of a freshly-installed appliance database (under four minutes here), which is what "3 of 3" waits on once the claim is gone.

## Tests

`agent/supervisor/tests/test_postgres_storage_reclaim.py` pins the contract (28 cases — the first 14 plus one per review finding: a forbidden, failing or unreachable PV read is an error with nothing deleted and only a missing PV falls back to the annotation, with the source recorded; a Cluster that names no primary, or has no status, defers everything and says why; the eviction tick reclaims with the Node still listed and the dead pod still reading Ready, a deferral names its node and the heartbeat carries it; a refused pod delete is an error with the claim still counted, a Terminating claim is pending not reclaimed, a vanished claim is not counted; the original 14: the live shape reclaims exactly the replica's claim then its pod; the current and the target primary are deferred, never deleted; a whole Postgres is one read; an eviction this tick scans past a stale status; a claim on a node that still exists — NotReady included — is left alone; WAL/tablespace claims go with their instance; unlabelled claims by CNPG naming; another cluster's, Redis's and network-storage claims are never touched; kubeapi failures and a refused delete are reported). `python -m pytest tests -q` in `agent/supervisor`: 402 passed (374 pre-existing + 28). No pre-existing failures.

CI mirror: agent tests as CI runs them (`pip install -e ".[dev]" && python -m pytest tests -q`, Python 3.14) green; the charts gate (`helm lint` + template + kubeconform) green with the rendered `spatium-supervisor-nodes` ClusterRole carrying `persistentvolumes: get`; gitleaks unchanged against the base (the same 4 pre-existing findings, delta 0). No `backend/`, `frontend/`, `appliance/` or migration paths are touched.

## Tradeoffs and known remaining work

* **The transient `spec.instances` 3 → 2 during a replace is left alone.** It is what `_committed_cp_count` implies and it also scales api/worker/frontend/redis, so changing it is a control-plane accounting decision, not a storage fix. It is worth its own look: when the dead node's instance is the primary at that moment, CNPG's scale-down (if it fires) sacrifices the *healthy* replica instead. This PR's sweep still converges that case — once the operator has failed over, the stranded claim is reclaimed and CNPG re-creates what is missing — but a replace should not have to pass through a primary-only window.
* `cluster/health` keeps `is_ha = control_plane_nodes > 1`; the `postgresql` workload row already reports `degraded 2/3` and is what the QA gate reads. Whether `is_ha` should also require the database to be whole is a UI/API contract question, not changed here.
* The slot-image mirror's claim has the same node affinity. Its Deployment references a fixed claim name, so reclaiming it would need the claim re-created from the chart; not done here, noted for the upgrade path.
* The replaced appliance row stays `approved`/`left` with its role assignments (the documented terminal state); nothing here changes the row.
* The QA drill that surfaced this graded `dead_node_replace` green on `nodes_ready >= 3` while CNPG sat at 2/3; the harness will grade the full gate predicate there (tracked on the QA side).

## Verification still to come

The QA side carries every open fix of ours, stacked on the next nightly's bytes, through a PostQA walk of both topologies (single node, then the three-node cluster). This fix rides that walk; the `dead_node_replace` drill's result on the stacked build will be posted here, as was done on #1053.

### DIFF
--- a/agent/supervisor/spatium_supervisor/heartbeat.py
+++ b/agent/supervisor/spatium_supervisor/heartbeat.py
@@ -1128,11 +1128,15 @@ def heartbeat_once(
         # tick are handed in as authoritative (the Node DELETE is
         # milliseconds old and the name can still be listed), so a
         # replica's claim goes on the eviction tick itself.
-        global _stranded_pending
         pg = k8s_api.reclaim_stranded_postgres_storage(
             evicted_nodes=set(evicted_now) | _stranded_pending
         )
-        _stranded_pending = _carry_stranded(_stranded_pending, evicted_now, pg)
+        # Mutated in place rather than rebound: the module-level set is the
+        # state, and a ``global`` rebinding reads as a never-read assignment
+        # to a static analyser (it is read on the next tick).
+        owed = _carry_stranded(_stranded_pending, evicted_now, pg)
+        _stranded_pending.clear()
+        _stranded_pending.update(owed)
         pg_reclaimed, pg_deferred, pg_err = pg.reclaimed, pg.deferred, pg.error
         if pg_reclaimed:
             log.info(
--- a/agent/supervisor/spatium_supervisor/heartbeat.py
+++ b/agent/supervisor/spatium_supervisor/heartbeat.py
@@ -1110,14 +1110,17 @@ def heartbeat_once(
         # has failed over. Every seed tick, not only the eviction one, so a
         # deferred primary and a supervisor restart both converge; one CR
         # read per tick while Postgres is whole.
-        pg_reclaimed, pg_deferred, pg_err = k8s_api.reclaim_stranded_postgres_storage(
-            evicted_nodes=evicted_now
-        )
+        pg = k8s_api.reclaim_stranded_postgres_storage(evicted_nodes=evicted_now)
+        pg_reclaimed, pg_deferred, pg_err = pg.reclaimed, pg.deferred, pg.error
         if pg_reclaimed:
             log.info(
                 "supervisor.heartbeat.postgres_storage_reclaimed",
                 pvcs=pg_reclaimed,
                 evicted=evicted_now,
+                # which read decided each claim was stranded: the PV's node
+                # affinity ("pv") or the scheduler's selected-node
+                # annotation ("annotation", only when the PV is gone)
+                affinity_source={name: pg.sources.get(name, "") for name in pg_reclaimed},
             )
         if pg_deferred:
             log.info(
--- a/agent/supervisor/spatium_supervisor/heartbeat.py
+++ b/agent/supervisor/spatium_supervisor/heartbeat.py
@@ -322,6 +322,21 @@ _last_peer_drift_at: float = 0.0
 # request; pruned once the backend drops the name from its evict list.
 _evicted_pending: set[str] = set()
 
+# #1058 — hostnames whose stranded CNPG claims the reclaim still owes (a
+# deferral: the instance was the primary, or the Cluster named none). Handed
+# back to the sweep as evicted names next tick so it scans past its
+# "Postgres is whole" early-out; cleared by the tick that settles them.
+_stranded_pending: set[str] = set()
+
+
+def _carry_stranded(previous: set[str], evicted_now: list[str], outcome) -> set[str]:
+    """PURE — what the next tick must force: after a clean sweep, exactly
+    what it left owed; after a failed one (kubeapi error), everything that
+    was owed plus this tick's evictions, so an outage never drops a node."""
+    if outcome.error is not None:
+        return set(previous) | {str(n) for n in evicted_now}
+    return set(outcome.pending_nodes)
+
 
 def _watchdog_check_due() -> bool:
     """First call always returns True (forces a probe within the
@@ -1109,8 +1124,15 @@ def heartbeat_once(
         # primary is deferred to the next tick, by which time the operator
         # has failed over. Every seed tick, not only the eviction one, so a
         # deferred primary and a supervisor restart both converge; one CR
-        # read per tick while Postgres is whole.
-        pg = k8s_api.reclaim_stranded_postgres_storage(evicted_nodes=evicted_now)
+        # read per tick while Postgres is whole. The names evicted THIS
+        # tick are handed in as authoritative (the Node DELETE is
+        # milliseconds old and the name can still be listed), so a
+        # replica's claim goes on the eviction tick itself.
+        global _stranded_pending
+        pg = k8s_api.reclaim_stranded_postgres_storage(
+            evicted_nodes=set(evicted_now) | _stranded_pending
+        )
+        _stranded_pending = _carry_stranded(_stranded_pending, evicted_now, pg)
         pg_reclaimed, pg_deferred, pg_err = pg.reclaimed, pg.deferred, pg.error
         if pg_reclaimed:
             log.info(
--- a/agent/supervisor/spatium_supervisor/heartbeat.py
+++ b/agent/supervisor/spatium_supervisor/heartbeat.py
@@ -1063,12 +1063,14 @@ def heartbeat_once(
         # backend clears the flag. Prune the stash to whatever the
         # backend still lists as pending (everything else is confirmed).
         evict_names = body_out.get("evict_node_names") or []
+        evicted_now: list[str] = []
         for name in evict_names:
             if name in _evicted_pending:
                 continue
             ok, evict_err = k8s_api.delete_node(str(name))
             if ok:
                 _evicted_pending.add(str(name))
+                evicted_now.append(str(name))
                 log.info("supervisor.heartbeat.node_evicted", node=name)
                 # #590 — the deleted node strands any local-path Redis PVC
                 # provisioned on it (node-affine PV): the replacement
@@ -1096,6 +1098,35 @@ def heartbeat_once(
                     "supervisor.heartbeat.node_evict_failed", node=name, error=evict_err
                 )
         _evicted_pending.intersection_update({str(n) for n in evict_names})
+
+        # #1058 — the deleted Node strands the CloudNativePG instance claim
+        # provisioned on it exactly as it strands Redis's, and the operator
+        # does NOT reclaim it: the re-created pod sits Pending on the dead
+        # node's PV affinity, Postgres runs 2 of 3 for good, and the
+        # replacement member never hosts an instance. Reclaim it the way
+        # the chart README says a human would — the claim, then the pod,
+        # and ONLY a replica's: an instance the Cluster still names as its
+        # primary is deferred to the next tick, by which time the operator
+        # has failed over. Every seed tick, not only the eviction one, so a
+        # deferred primary and a supervisor restart both converge; one CR
+        # read per tick while Postgres is whole.
+        pg_reclaimed, pg_deferred, pg_err = k8s_api.reclaim_stranded_postgres_storage(
+            evicted_nodes=evicted_now
+        )
+        if pg_reclaimed:
+            log.info(
+                "supervisor.heartbeat.postgres_storage_reclaimed",
+                pvcs=pg_reclaimed,
+                evicted=evicted_now,
+            )
+        if pg_deferred:
+            log.info(
+                "supervisor.heartbeat.postgres_storage_reclaim_deferred",
+                instances=pg_deferred,
+                reason="instance is the current or target primary; retrying next tick",
+            )
+        if pg_err:
+            log.warning("supervisor.heartbeat.postgres_storage_reclaim_failed", error=pg_err)
 
         # #590 — cluster DNS must survive a node loss. k3s's bundled
         # CoreDNS is a single replica that deterministically sits on the
--- a/agent/supervisor/spatium_supervisor/heartbeat.py
+++ b/agent/supervisor/spatium_supervisor/heartbeat.py
@@ -1150,6 +1150,12 @@ def heartbeat_once(
                 instances=pg_deferred,
                 reason=pg.deferred_reason,
             )
+        if pg.terminating:
+            log.info(
+                "supervisor.heartbeat.postgres_storage_reclaim_pending",
+                pvcs=pg.terminating,
+                reason="deleted on an earlier tick; pvc-protection is waiting for the pod",
+            )
         if pg_err:
             log.warning("supervisor.heartbeat.postgres_storage_reclaim_failed", error=pg_err)
 
--- a/agent/supervisor/spatium_supervisor/heartbeat.py
+++ b/agent/supervisor/spatium_supervisor/heartbeat.py
@@ -1126,7 +1126,7 @@ def heartbeat_once(
             log.info(
                 "supervisor.heartbeat.postgres_storage_reclaim_deferred",
                 instances=pg_deferred,
-                reason="instance is the current or target primary; retrying next tick",
+                reason=pg.deferred_reason,
             )
         if pg_err:
             log.warning("supervisor.heartbeat.postgres_storage_reclaim_failed", error=pg_err)
--- a/agent/supervisor/spatium_supervisor/k8s_api.py
+++ b/agent/supervisor/spatium_supervisor/k8s_api.py
@@ -667,6 +667,10 @@ class PostgresReclaim:
     # hands them back as ``evicted_nodes`` next tick so the scan runs again
     # even while the Cluster reads whole
     pending_nodes: set[str] = field(default_factory=set)
+    # claims an earlier tick already deleted that are still Terminating
+    # (pvc-protection waits for the consumer pod) — reported as pending,
+    # never counted as reclaimed again
+    terminating: list[str] = field(default_factory=list)
 
     def __iter__(self):
         # ``reclaimed, deferred, err = ...`` keeps working for callers that
@@ -791,11 +795,13 @@ def reclaim_stranded_postgres_storage(
     stranded: dict[str, list[str]] = {}
     sources: dict[str, str] = {}
     pinned_to: dict[str, set[str]] = {}
+    already_terminating: set[str] = set()
     for pvc in items:
         instance = _cnpg_instance_of(pvc, cluster_name)
         if instance is None:
             continue
-        name = str((pvc.get("metadata") or {}).get("name") or "")
+        meta = pvc.get("metadata") or {}
+        name = str(meta.get("name") or "")
         pinned, source, pin_err = _pvc_pinned_hostnames(pvc)
         if pin_err:
             return PostgresReclaim(error=f"{name}: {pin_err}")
@@ -804,9 +810,12 @@ def reclaim_stranded_postgres_storage(
         stranded.setdefault(instance, []).append(name)
         sources[name] = source
         pinned_to.setdefault(instance, set()).update(pinned)
+        if meta.get("deletionTimestamp"):
+            already_terminating.add(name)
 
     reclaimed: list[str] = []
     deferred: list[str] = []
+    terminating: list[str] = []
     deferred_reason = ""
     for instance, claims in sorted(stranded.items()):
         if primary_unknown:
@@ -818,27 +827,48 @@ def reclaim_stranded_postgres_storage(
             deferred_reason = "instance is the current or target primary; retrying next tick"
             continue
         for name in sorted(claims):
+            if name in already_terminating:
+                # Deleted on an earlier tick; pvc-protection is holding it
+                # until its pod is gone. Not ours to count again — the first
+                # cut re-reported it as "reclaimed" every tick for ever
+                # (review, point 4). The pod delete below is what it waits on.
+                terminating.append(name)
+                continue
             try:
                 status, resp = _request(
                     "DELETE", f"{base}/persistentvolumeclaims/{quote(name)}"
                 )
             except RuntimeError as exc:
-                return PostgresReclaim(reclaimed, deferred, str(exc), sources, deferred_reason)
+                return PostgresReclaim(reclaimed, deferred, str(exc), sources,
+                                       deferred_reason, terminating=terminating)
             if status not in (200, 202, 404):
-                return PostgresReclaim(reclaimed, deferred, f"kubeapi status {status}: {resp[:200]!r}", sources, deferred_reason)
-            reclaimed.append(name)
+                return PostgresReclaim(reclaimed, deferred,
+                                       f"kubeapi status {status}: {resp[:200]!r}",
+                                       sources, deferred_reason, terminating=terminating)
+            if status != 404:  # 404: vanished between the list and the delete — not ours
+                reclaimed.append(name)
         # The pod the operator re-created against the stranded claim is
         # unscheduled, so a plain DELETE removes it at once and releases
         # pvc-protection; CNPG then sees a missing instance and joins a
-        # fresh one (pg_basebackup from the primary) on a live node.
+        # fresh one (pg_basebackup from the primary) on a live node. A
+        # refused delete leaves the claim Terminating under pvc-protection
+        # and CNPG unable to re-create a claim of that name, so it is an
+        # error, not a shrug (review, point 4).
+        pod_path = f"{base}/pods/{quote(instance)}"
         try:
-            _request("DELETE", f"{base}/pods/{quote(instance)}")
-        except RuntimeError:
-            pass  # the pod may already be gone; the claim reclaim is what matters
+            status, resp = _request("DELETE", pod_path)
+        except RuntimeError as exc:
+            return PostgresReclaim(reclaimed, deferred, f"pod {instance}: {exc}", sources,
+                                   deferred_reason, terminating=terminating)
+        if status not in (200, 202, 404):
+            return PostgresReclaim(reclaimed, deferred,
+                                   f"pod {instance}: kubeapi status {status}: {resp[:200]!r}",
+                                   sources, deferred_reason, terminating=terminating)
     pending = set()
     for instance in deferred:
         pending |= pinned_to.get(instance, set())
-    return PostgresReclaim(reclaimed, deferred, None, sources, deferred_reason, pending)
+    return PostgresReclaim(reclaimed, deferred, None, sources, deferred_reason, pending,
+                           terminating)
 
 
 # #272 — durable control-plane state via k3s HelmChartConfig.
--- a/agent/supervisor/spatium_supervisor/k8s_api.py
+++ b/agent/supervisor/spatium_supervisor/k8s_api.py
@@ -660,6 +660,9 @@ class PostgresReclaim:
     deferred: list[str] = field(default_factory=list)
     error: str | None = None
     sources: dict[str, str] = field(default_factory=dict)
+    # why ``deferred`` is non-empty: the instance is the current or target
+    # primary, or the Cluster names no primary at all
+    deferred_reason: str = ""
 
     def __iter__(self):
         # ``reclaimed, deferred, err = ...`` keeps working for callers that
@@ -701,7 +704,8 @@ def reclaim_stranded_postgres_storage(
     as ``currentPrimary`` or ``targetPrimary`` is deferred — the caller
     retries on its next tick, by which time the operator has failed over
     (the primary's pod is gone with the node) and the same instance is a
-    stranded replica.
+    stranded replica. A Cluster that names no primary at all defers
+    everything: an unknown primary is not "no primary".
 
     Bounded and idempotent: one GET of the Cluster CR when Postgres is
     whole (``readyInstances >= instances``) and nothing was evicted this
@@ -736,6 +740,13 @@ def reclaim_stranded_postgres_storage(
     primaries = {
         str(cr_status.get(key) or "") for key in ("currentPrimary", "targetPrimary")
     } - {""}
+    # A Cluster that names NO primary (no status yet, a status wiped by an
+    # operator restart, a bootstrap in progress) is not one where every
+    # instance is a replica — it is one where the primary is unknown.
+    # Deleting anything on that reading could be the primary's PGDATA
+    # (review of the first cut, point 2), so nothing is; the next tick,
+    # with a status, decides. The deferral costs one tick.
+    primary_unknown = not primaries
 
     try:
         status, resp = _request("GET", "/api/v1/nodes")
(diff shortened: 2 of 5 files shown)


mzac (commented):
Reviewed at `0c70a2b` against a clean worktree. Suite reproduces green — **388 passed** on py3.14, matching your count.

The diagnosis is right and the shape of the fix is right: CNPG genuinely does not reclaim a node-affine claim whose node is gone, replica-only is the correct constraint, running every tick rather than once at eviction is the right call, and the 14 test cases each name a real property. This fixes a real "Postgres is 2 of 3 forever".

@amoona6 — could you validate the five points below? Four are reproduced against this branch by stubbing `k8s_api._request`; the first is a chart/RBAC read. My own read is that #1 (inline, line 630) and the `live -= forced` one-liner (line 714) are worth folding in before merge — one is provably not doing what the PR says, and the other makes the tested guard unreachable in production. The rest are defensible as follow-ups. Happy to be wrong on any of them.

---

### The `persistentvolumes` read is not granted by RBAC

This half is in a file outside the diff, so it can't be an inline comment: `charts/spatiumddi-appliance/templates/supervisor-rbac.yaml`.

The supervisor runs as an in-cluster pod — `serviceAccountName: spatium-supervisor`, `automountServiceAccountToken` unset so it defaults true, and `spatiumddi-firstboot` renders `supervisor.enabled: true` — so `_resolve_config()` takes the in-cluster branch and gets exactly that RBAC rather than the hostPath kubeconfig. The complete grant list there is helmcharts/helmchartconfigs, pods/services/events/configmaps, persistentvolumeclaims, pods delete, cnpg clusters, nodes, etcdsnapshotfiles, deployments. There is no `persistentvolumes`; `grep -rn persistentvolumes charts/ k8s/` finds one grant in the whole tree and it belongs to kube-state-metrics.

So `GET /api/v1/persistentvolumes/{name}` answers **403** on every appliance, and `_pvc_pinned_hostnames` falls through to the `selected-node` annotation.

Consequences:

1. Every production reclaim decides on the annotation, not on PV affinity — including the one in your drill. On local-path the two name the same node and produce an identical log line, so the hardware run cannot distinguish them: it is evidence the *fallback* works.
2. `test_network_storage_is_never_stranded` cannot hold. Re-running that test's exact scenario with the PV GET answering 403 returns `(['spatium-control-spatiumddi-postgresql-3'], [], None)` and issues `DELETE .../persistentvolumeclaims/...-3` then `DELETE .../pods/...-3` — the claim the test exists to protect, deleted, with `err is None`. Latent on the appliance (local-path only), but a CSI class with `volumeBindingMode: WaitForFirstConsumer` also sets that annotation.
3. The 403 is swallowed, which is at odds with the docstring's "kubeapi errors are surfaced, never guessed around".

Fix is `persistentvolumes: ["get"]` on the cluster-scoped `spatium-supervisor-nodes` ClusterRole, plus distinguishing 403 from 404 at line 630.

---

Mechanically the PR is `MERGEABLE` with 33/33 checks green, but `BEHIND` main — it predates #1053.

agent/supervisor/spatium_supervisor/k8s_api.py
```diff
@@ -564,6 +568,206 @@ def reclaim_stranded_redis_storage(
...
+    the scheduler's ``selected-node`` annotation — what the Redis reclaim
+    keys on — is the fallback when the PV cannot be read. A readable PV
+    with no hostname affinity is network storage: nothing pins it."""
+    volume = str((pvc.get("spec") or {}).get("volumeName") or "")
+    if volume:
+        try:
+            status, resp = _request("GET", f"/api/v1/persistentvolumes/{quote(volume)}")
+        except RuntimeError:
+            status, resp = 0, b""
+        if status == 200:
```
mzac, line 640: A 403 is indistinguishable from a 404 here: `if status == 200` sends both to the annotation fallback with no error reported. Per the review body the supervisor's ServiceAccount has no `persistentvolumes` grant, so this is the branch every appliance actually takes — the PV-affinity path never runs in production, and the "a readable PV with no hostname affinity is network storage" guard goes with it.

Worth splitting: 404 (PV genuinely gone) → annotation fallback as today; 403 / 5xx → return the error so it surfaces as `postgres_storage_reclaim_failed` instead of silently downgrading the signal.

agent/supervisor/spatium_supervisor/k8s_api.py
```diff
@@ -564,6 +568,206 @@ def reclaim_stranded_redis_storage(
...
+        cr = json.loads(resp)
+        spec = cr.get("spec") or {}
+        cr_status = cr.get("status") or {}
+        want = int(spec.get("instances") or 0)
+        ready = int(cr_status.get("readyInstances") or 0)
+    except (ValueError, TypeError, AttributeError):
+        return [], [], "unparseable Cluster CR"
+    if ready >= want and not forced:
+        return [], [], None
+    primaries = {
```
mzac, line 751: This fails open when the Cluster names no primary. `- {""}` strips the empty strings, so a CR whose status carries neither `currentPrimary` nor `targetPrimary` yields an empty set, nothing is deferred, and a stranded **primary's** claim is deleted.

Reproduced with `{"spec": {"instances": 3}, "status": {"readyInstances": 2}}` — returns `(['spatium-control-spatiumddi-postgresql-3'], [], None)` and deletes the claim plus the pod. A CR with no `status` at all lands in the same place: `ready` defaults to 0, `want` is 3 from spec, so `0 >= 3` is false and the scan runs with no primary protection.

The whole safety argument — in the docstring and in the CHANGELOG — is "only a replica's claim is ever deleted", and it rests entirely on this field. When the field can't be read the code reads it as "not the primary" rather than "can't tell". Since the deferral path already exists and costs one tick, an unnamed primary seems like it should defer. This is the only finding whose consequence is deleting a primary's PGDATA, which is why I'd rather not leave it to a follow-up.

agent/supervisor/spatium_supervisor/k8s_api.py
```diff
@@ -564,6 +568,206 @@ def reclaim_stranded_redis_storage(
...
+    } - {""}
+
+    try:
+        status, resp = _request("GET", "/api/v1/nodes")
+    except RuntimeError as exc:
+        return [], [], str(exc)
+    if status != 200:
+        return [], [], f"kubeapi status {status}: {resp[:200]!r}"
+    try:
+        live = {
```
mzac, line 769: `forced` is used in exactly one place — the `ready >= want` early-out — and is never subtracted from `live`. So on the eviction tick the Node DELETE is milliseconds old, the name can still come back in this list, `pinned & live` is non-empty, and the tick is a silent no-op.

I think your own journal shows this happening:

```
21:11:23.988Z node_evicted                 node=ddipg-member-2
21:11:24.414Z redis_storage_reclaimed      node=ddipg-member-2
21:21:55.493Z postgres_storage_reclaimed   pvcs=[...] evicted=[]
```

Redis reclaims on the eviction tick because it matches the annotation against the node name it was handed and never consults the Node list. Postgres has no such anchor, so it converged 10m31s later on a tick that evicted nothing — `evicted=[]`. The PR attributes that gap to the transient `spec.instances` 3→2, but `forced` is precisely what is supposed to bypass that, so either way the fast path did not fire.

`live -= forced` before the scan gives Postgres the Redis path's determinism and closes the 2-of-3 window. Note `test_an_eviction_this_tick_forces_the_scan_past_a_stale_status` can't catch this — it scripts a node list that already excludes the evicted node, so the race isn't in scope for it.

agent/supervisor/spatium_supervisor/k8s_api.py
```diff
@@ -564,6 +568,206 @@ def reclaim_stranded_redis_storage(
...
+                return reclaimed, deferred, str(exc)
+            if status not in (200, 202, 404):
+                return reclaimed, deferred, f"kubeapi status {status}: {resp[:200]!r}"
+            reclaimed.append(name)
+        # The pod the operator re-created against the stranded claim is
+        # unscheduled, so a plain DELETE removes it at once and releases
+        # pvc-protection; CNPG then sees a missing instance and joins a
+        # fresh one (pg_basebackup from the primary) on a live node.
+        try:
+            _request("DELETE", f"{base}/pods/{quote(instance)}")
```
mzac: The status is discarded here — only `RuntimeError` is caught, so any HTTP failure is invisible. With the pod DELETE answering 500 the function returns `(['spatium-control-spatiumddi-postgresql-3'], [], None)`: reported as a clean reclaim (reproduced).

That state is worse than it looks. The claim is already gone but the pod still references it, so the PVC sits in `Terminating` under `kubernetes.io/pvc-protection` and CNPG can't create a replacement claim of that name. And because a Terminating PVC is still returned by the list call while DELETE answers 200/404, the next tick re-detects it and appends it to `reclaimed` again — two consecutive calls both return `(['spatium-control-spatiumddi-postgresql-3'], [], None)`, also reproduced. The heartbeat then logs `postgres_storage_reclaimed` at INFO every 30 s indefinitely with `pg_err` never set, which reads as repeated success rather than a wedged instance.

Capturing the status and folding a non-`(200, 202, 404)` into the returned error would cover it; appending to `reclaimed` only on a real 200/202 would also keep an idempotent re-run quiet. (The Redis reclaim has the same shape, but this is the function whose docstring makes the PVC-then-pod ordering the contract.)

agent/supervisor/spatium_supervisor/k8s_api.py
```diff
@@ -564,6 +568,206 @@ def reclaim_stranded_redis_storage(
...
 
+# #1058 — CloudNativePG instance storage stranded on an evicted node.
+#
+# CNPG's own labels on the claims and pods it creates (docs: "Labels and
+# annotations"). The name fallback below follows the naming CNPG documents:
+# ``<cluster>-<ordinal>`` for the PGDATA claim, with ``-wal`` / ``-tbs-<name>``
+# suffixes for WAL and tablespace claims.
+_CNPG_CLUSTER_LABEL = "cnpg.io/cluster"
+_CNPG_INSTANCE_LABEL = "cnpg.io/instanceName"
+_CNPG_DEFAULT_CLUSTER = "spatium-control-spatiumddi-postgresql"
```
mzac, line 579: Minor: this is now the second spelling of the same string — `patch_cnpg_instances` carries it inline as a parameter default at k8s_api.py:1681. Same CR, same namespace, and a release rename would have to find both. Worth pointing that default at this constant in the same edit.

agent/supervisor/spatium_supervisor/k8s_api.py
amoona6: Confirmed — the ClusterRole had no `persistentvolumes` grant, so this branch was the one every appliance took and the drill's reclaim decided on the annotation. Addressed in 3803299b: the grant on `spatium-supervisor-nodes`, and the read now returns (hostnames, source, error) — 200 → PV affinity, 404 → annotation fallback, 403 / 5xx / transport → `postgres_storage_reclaim_failed` with nothing deleted; `affinity_source` per claim in the journal. Live on the amended chart: can-i → yes, the PV GET with the supervisor's token → 200, the reclaim line reads `affinity_source=pv`.

agent/supervisor/spatium_supervisor/k8s_api.py
amoona6: Confirmed, and agreed this is the one to fix here. Addressed in 5013563b: a Cluster that names neither `currentPrimary` nor `targetPrimary` — or has no status at all — defers every stranded instance, and the deferral says why in the journal. Three tests pin it (no primary named; no status; the reason text).

agent/supervisor/spatium_supervisor/k8s_api.py
amoona6: Confirmed, and your reading of the journal is right — with a correction to ours: both halves are true. Our first-round node list 1.5 s after `node_evicted` still carried the evicted member (silent no-op, no postgres line on that tick), and every later tick early-outed on the scaled-down spec until the replacement joined; the PR had blamed the whole 10 min 31 s on #1059. Addressed in 240955da: `live -= forced` (the names the seed just evicted are authoritative) and a deferral carried to the next tick by the heartbeat. Live on the amended build the reclaim landed 0.88 s after the Node deletion, on the eviction tick, with the 3→2 scale requested 0.2 s earlier and no longer in the way.

agent/supervisor/spatium_supervisor/k8s_api.py
amoona6: Confirmed. Addressed in e22a29ea: the pod DELETE's status is folded into the returned error (200/202/404 ok); a claim with a `deletionTimestamp` is treated as terminating → `postgres_storage_reclaim_pending`, never re-deleted and never re-reported as "reclaimed"; a 404 on the claim delete is not counted as a reclaim. Tests cover a refused pod delete, a Terminating claim on the next tick, and a claim that vanished before the delete.

agent/supervisor/spatium_supervisor/k8s_api.py
amoona6: Done in 52c9284e: `patch_cnpg_instances` defaults to `_CNPG_DEFAULT_CLUSTER`.

agent/supervisor/spatium_supervisor/heartbeat.py
```diff
@@ -1097,6 +1114,51 @@ def heartbeat_once(
...
+        # deferred primary and a supervisor restart both converge; one CR
+        # read per tick while Postgres is whole. The names evicted THIS
+        # tick are handed in as authoritative (the Node DELETE is
+        # milliseconds old and the name can still be listed), so a
+        # replica's claim goes on the eviction tick itself.
+        global _stranded_pending
+        pg = k8s_api.reclaim_stranded_postgres_storage(
+            evicted_nodes=set(evicted_now) | _stranded_pending
+        )
+        _stranded_pending = _carry_stranded(_stranded_pending, evicted_now, pg)
```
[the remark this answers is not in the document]
amoona6: The set is read on every tick (`evicted_nodes=set(evicted_now) | _stranded_pending`) and was rebound through `global` for the next one, which reads as a never-read assignment to the analyser. Fixed in e508b945: the module-level set is mutated in place (cleared and refilled from the pure `_carry_stranded`), no behaviour change; the supervisor suite is 402 passed as CI runs it.

Issue #882, opened 2026-08-21
Agents never read the previous.json they write — auto-revert to last-known-good on apply failure


mzac:
From the 2026-08-21 idea pass (the "last known good configuration" half of the config-rollback idea; control-plane snapshots are tracked separately).

## The gap

Non-negotiable #5 (agents cache last-known-good config and run from cache when the control plane is unreachable) is implemented for the *unreachable* case — but not for the *bad config* case, and the code that looks like it handles it is dead:

- Both agents write an N-1 copy on every config change: `write_config` does `current.replace(previous)` — `agent/dns/spatium_dns_agent/cache.py:71-75`, `agent/dhcp/spatium_dhcp_agent/cache.py:89-96` (also the LG agent keeps `current.json`).
- **`previous.json` is written but never read.** Grep across `agent/dns/` and `agent/dhcp/` for `previous`/`rollback`/`revert` finds no consumer.

So a bad-but-parseable bundle (renders to config the daemon rejects, or renders fine but the daemon fails on it) overwrites the last-known-good and the agent has nothing to fall back to; it will keep re-applying the same broken ETag.

## Proposal

1. **Apply is transactional per bundle**: render → validate (`named-checkconf` for BIND9; Kea config-test / dry-run where available) → apply → health-confirm (daemon up + answering after N seconds). Any failure ⇒ **revert to `previous.json`**, re-render, re-apply, and verify the daemon is healthy on the old config.
2. **Poison-pill guard**: remember the failed ETag and don't re-apply it on the next long-poll cycle; resume normal behavior when the server presents a different ETag. (The 12 s wake tick + 2 s fallback poll otherwise turns a bad bundle into a crash loop.)
3. **Report upward**: agent includes `config_apply_status {ok | reverted, failed_etag, error}` on its next report; control plane persists it on the server row, surfaces a rose chip on the server list + zone/scope detail, and fires an alert rule (`agent_config_rejected`) — otherwise a revert silently masks that new config never converged.
4. **Depth**: keep exactly N-1 (current behavior) — the control plane is the source of truth for anything older; deeper history on-agent buys little.
5. Cover DNS, DHCP, and the LG agent's gobgpd config; the supervisor's host-config runners already have per-runner status sidecars and mostly validate-before-swap (the firewall runner is the reference: staged file + `nft -c` + revert timer) — audit the newer runners against that same bar as part of this issue.

[...]

PR #1061, opened 2026-09-11
fix(dns-agent): a deferred daemon start is a wait, not a death (#1056)

### LINKED ISSUE #1056: dns agent: a bind pod started before its first config is rendered exits 2 one second after 'startup deferred', so every freshly joined member's bind pod crash-loops until its config arrives
**What happens.** On a member that has just joined the cluster, the `dns-bind9` DaemonSet pod starts before the control plane has handed the agent its first config. `Bind9Driver.start_daemon()` sees no `named.conf` and returns without starting `named` (`agent/dns/spatium_dns_agent/drivers/bind9.py:1866-1869`, logging `named_conf_missing_startup_deferred`). One second later the supervisor's main loop asks `driver.daemon_running()`, gets `False`, logs `dns_daemon_exited` and returns 2 (`agent/dns/spatium_dns_agent/supervisor.py:129-136`: "if the daemon dies, we exit non-zero so the orchestrator restarts the container"). The container exits with code 2 and kubelet back-off-restarts it. The pod keeps crash-looping, with growing back-off, until the first config lands and `start_daemon()` has something to start.

The previous container's whole log, from a member that joined on `nightly-2026.09.10` + PR #1053 (2026-09-11):

```
15:24:32Z {"event": "named_conf_missing_startup_deferred", "service": "spatium-dns-agent", "level": "warning"}
15:24:32Z {"event": "dns_query_log_shipper_starting", ...}
15:24:32Z {"event": "dns_ingest_worker_starting", ...}
15:24:33Z HTTP Request: POST https://<seed>/api/v1/dns/agents/heartbeat "HTTP/1.1 200 OK"
15:24:33Z {"driver": "bind9", "event": "dns_daemon_exited", "service": "spatium-dns-agent", "level": "error"}
```

`kubectl describe` on that pod: `Last State: Terminated, Reason: Error, Exit Code: 2, Started 15:24:31, Finished 15:24:33`, `Restart Count: 2`, then `Ready: True` once the config had arrived (running since 15:24:46). Events show the same `Back-off restarting failed container dns-bind9` on every bind pod the DaemonSet created on the two new members during formation (four pods, two of them deleted and re-created by a DaemonSet rollout in the same minute).

**How often.** Every cluster formation we have measured on a healthy join: +1 restart on one walk (2026-09-04), +1 on our 2026-09-09 verification walk, +2 on the 2026-09-10 one. It is deterministic in kind and varies only in how many back-off rounds fit before the config arrives. Nothing else restarts on those formations (api, worker, kea, k3s all 0).

**Why it matters.** The member is not serving DNS yet, so no query is lost — but (1) the back-off delays the member's first DNS service by the sum of the back-off intervals, which grow with every round; (2) a restart count that is never zero on a healthy join hides the restarts that do matter (our formation gate counts restarts in the window for exactly that reason, and this is the one thing that keeps it from reading 0 on a good cluster); (3) "startup deferred" says the agent will wait, and it does not — the log promises one behaviour and the process does the other.

**Expected.** A deferred start should be a wait, not a death: gate the `daemon_running()` check on the daemon having been started at least once (or on the config existing), so the supervisor idles until the first sync renders `named.conf` and then starts `named` — the daemon-death exit stays for a daemon that actually died. The same shape may apply to the DHCP agent's supervisor if it shares the loop; we only observed bind.

Evidence on request: the pod's `describe`, both container logs, the DaemonSet events and the formation timeline.

amoona6:
Fixes #1056 — the `dns-bind9` pod on a freshly joined cluster member exited 2 one second after logging `named_conf_missing_startup_deferred`, and kubelet back-off-restarted it until its first config happened to land inside the 1 s window. The same supervisor loop also reported every DaemonSet rollout as a crash (exit 2 on a requested stop); that is the second commit. The review round added the race guards on both crash exits, kept the wait visible (re-logged on a backoff, `degraded` in the heartbeat), gave the umbrella chart's DNS agents the liveness probe the appliance chart already had, declared `daemon_pid` once, added the CHANGELOG entry, and measures the wait from the loop's start.

Root-caused live on nested 3-node QA clusters (three formations with the same signature: +1, +1, +2 restarts), fixed, and proven three times on the same base: the trigger fired by hand on a formed cluster, a full formation graded by the QA harness with the trigger inside its window, and — after the review — the amended stack on a third formation plus a stuck-agent drill for the bounded wait.

## 1. A never-launched daemon was read as a dead one

`supervisor.run()` calls `driver.start_daemon()` once at boot, then every second treats `driver.daemon_running() == False` as "the daemon died, exit 2 so the orchestrator restarts us". But `start_daemon()` is allowed to return *without* a daemon: `Bind9Driver.start_daemon` defers when no `named.conf` has been rendered yet and leaves the launch to `swap_and_reload`, which the sync loop reaches once the control plane hands over the first bundle (`PowerDNSDriver` defers the same way). The first bind pod on a freshly joined member always boots deferred — its hostPath state dir is empty and its bundle is not built yet — so the process exited at +1.0 s whenever the first bundle took longer than that.

The previous container's whole log on the member that counted +2 (2026-09-11, `nightly-2026.09.10` + #1053):

```
15:24:32.232Z {"event": "named_conf_missing_startup_deferred", "service": "spatium-dns-agent", "level": "warning"}
15:24:32.241Z {"event": "dns_query_log_shipper_starting", "path": "/var/log/named/queries.log", "level": "info"}
15:24:32.244Z {"event": "dns_ingest_worker_starting", "interval": 180.0, "level": "info"}
15:24:33.205Z HTTP Request: POST https://<seed>/api/v1/dns/agents/heartbeat "HTTP/1.1 200 OK"
15:24:33.245Z {"driver": "bind9", "event": "dns_daemon_exited", "service": "spatium-dns-agent", "level": "error"}
```

`kubectl describe`: `Last State: Terminated, Reason: Error, Exit Code: 2, Started 15:24:31, Finished 15:24:33`, `Restart Count: 2`, `Back-off restarting failed container dns-bind9`. The `GET /api/v1/dns/agents/config` never completed inside the container's life. Pods that read 0 on the same formation were the ones whose predecessor had already cached a bundle on the hostPath: `SyncLoop.__init__` applies it — and launches `named` — before the supervisor's own `start_daemon()` runs, so nothing defers there.

**The fix.** The liveness check is armed by the launch, not by the boot:

- `DriverBase.daemon_launched()` — true once the driver has spawned or adopted its daemon. `daemon_pid` is set at exactly those points (the spawn and the adopt in `start_daemon`, the system-wide look-up in `daemon_running`) and never cleared, so it is the tell. It is declared once, on `DriverBase`; the three subclass declarations are gone (review point 4 — the first version had added a fourth copy while calling it a move).
- `supervisor.run()` exits 2 only when `daemon_launched()` and not `daemon_running()`. While nothing has been launched it waits. `daemon_running()` is still evaluated first, so the #704 adopt path is unchanged.

Why the pid rather than "the config exists": `swap_and_reload` promotes `rendered.new → rendered` *before* it spawns, so a config-on-disk predicate has a window in which the main thread sees config and no daemon — a rarer version of the same bug. A pid exists only once a process does.

**What bounds the wait.** The liveness probes — on both charts now. The appliance chart's `dns-bind9` / `dns-powerdns` / `dns-technitium` already had `tcpSocket :53`, 30 s initial delay, 30 s period; the umbrella chart's `dns-agent` StatefulSet had a readiness probe only, and a readiness probe never restarts a pod (review point 1 — the first version claimed the bound for "the chart" and it held for one of the two). The umbrella chart's DNS agents get the same probe in this PR, so a bundle that never arrives (wrong group, revoked key, a control plane that answers `/heartbeat` but renders nothing) is a restart every ~2 min with the reason in the agent's log, on either chart, instead of a silent NotReady. A Compose deployment without a probe waits, heartbeating — and says so (next section).

## 2. A requested stop was reported as a death — wherever it lands

On SIGTERM the handler sets `stopping` and stops every worker thread, and a DaemonSet rollout may take the daemon in the same instant. The loop's next tick ran its two liveness checks first, found the threads it had just stopped (or the daemon the rollout killed) dead, logged `dns_agent_thread_died` / `dns_daemon_exited` and returned 2 — every rollout ended the old container with exit 2. Seen live on the seed when the serving roles were mirrored onto the members: kubelet `StopContainer … with signal terminated` at 21:20:52.96Z, the `dns-bind9` container's exit event 0.7 s later with `exit_status:2`; the `dhcp-kea` container replaced in the same second exited 0. The rollout that triggers it on every fresh cluster is the chart's own: once the serving roles land, the DaemonSet template is re-rendered with the external control-plane URL and every agent pod on every node is replaced.

**The fix.** A stop requested during the tick takes the designed exit before either check runs; and because Python runs signal handlers between bytecodes, both crash exits also require that no stop has been requested (review point 2 — a stop landing *inside* `daemon_running()` or between the two checks still reached them; reproduced with the fixture, rc 2 on both paths). The residual window is the few bytecodes between that test and the `return`. A worker thread that dies while the agent is *not* stopping still exits 2.

**What this does and does not cover.** Every container a rollout stops *once the agent is running* exits 0. A container stopped in its first few hundred milliseconds — before the entrypoint has handed off to the agent and the agent has installed its handlers, which happens after registration, driver construction and thread start — dies by the signal, as any process without a handler does: 143. Seen once in the review round, when a revision-2 pod's container creation raced the revision-3 rollout (the pod was already Terminating when kubelet started the container; 282 ms of life, no log line). That pre-handler window — the entrypoint shell, interpreter start-up, registration — is outside the loop's reach; closing it would need a bootstrap-time handler that turns SIGTERM into a clean exit, and it can never cover the shell. Not in this PR.

## 3. The wait is visible

With the crash loop gone, the waiting state was the only signal a stuck agent produced, and it was logged once, at t=1 s (review point 3). Now: `dns_daemon_start_deferred_waiting` is re-logged on a backoff — the first tick, then doubling to 32 s, then every minute — each line carrying `waited_s`, so a `kubectl logs --tail` an hour into the wait still shows the state without a line a second; and the heartbeat carries `daemon: {"status": "degraded", "reason": "start deferred, no bundle yet"}` for as long as the wait lasts, cleared to `ok` the moment the daemon is up — only if it is still the supervisor's own, so a degraded verdict `sync.py` set meanwhile (a failed apply) is left alone. The control plane discards the heartbeat's `daemon` field today (`agents.py` declares it, nothing reads it — the #882 class); the signal is there for when it does.

Nested 3-node clusters (seed + two members) on the QA Proxmox fleet; the formation gate counts every pod's `restartCount` growth across the window (threshold 0). The trigger is a member's *first* bind pod with an empty state dir.

**A. Trigger fired by hand on a formed cluster (commit 1)** — the seed's serving roles were mirrored onto both members with the same product API the gate uses:

| | Unfixed control (same base, member-2) | Fixed (member-2) | Fixed (member-1) |
|---|---|---|---|
| deferral → first bundle | `GET /config` never completed in the container's 1.0 s life | deferred 21:20:42.767 → `dns_daemon_start_deferred_waiting` :43.809 → `GET …/config 200` :43.864 → `named_started` :43.894 → `dns_daemon_launched_after_deferred_start waited_s 1.0` — **first bundle +1.10 s, past the old window** | deferred :43.062 → `GET 200` :43.746 → `named_started` :43.770 (+0.71 s) |
| `dns_daemon_exited` | at +1.012 s, twice | none | none |
| Restart Count / Last State | 2 / Terminated exit 2 | **0** / none | **0** / none |
| pod scheduled → Ready | two back-off rounds; serving container at +25 s | Ready at **+15 s** | Ready at +13 s |
| `version.bind CH TXT` from the node | (not measured) | `"9.20.27"` from all three node addresses | same |
| formation gate `pod_restarts` over the window | **fail** — `2 pod restart(s): dns-bind9-<member-2> +2` | **pass** — `0 pod restart(s) within the threshold of 0, 0 OOMKilled` | — |

**B. A full formation graded by the harness with the trigger inside its window (commits 1+2)** — the gate itself mirrored the roles during formation:

| check | result |
|---|---|
| `pod_restarts` | **pass** — `0 pod restart(s) within the threshold of 0, 0 OOMKilled` |
| `serving_roles_all_nodes` | pass — `roles ['dhcp', 'dns-bind9'] mirrored from the seed onto 2/2 member(s) …; dns agents 3/3, dhcp agents 3/3 registered` |
| `sites_serving_all_nodes` | pass — `3/3 nodes each answer the site's DHCP DISCOVER and DNS query` |
| `cluster_formation` / `_unrepaired` · `k3s_restarts` · `gate_predicate` | `2 members paired+approved+promoted (first try)` · `+0/+0/+0` · `3/3 ready, HA, CNPG=3` |
| members' first bind pods | member-2: deferred 22:30:04.99 → `named_started` :06.00 (+1.01 s); member-1: deferred :09.74 → `named_started` :10.61 (+0.87 s); no previous container anywhere |
| the rollout (commit 2) | six agent containers SIGTERMed on all three nodes at 22:30:15–49Z — **all exit 0**; member-1's log: `dns_agent_signal_received` 22:30:14.86 → `dns_agent_exiting` 22:30:15.76 (the same rollout ended the unfixed container with exit 2) |

**C. The amended stack on a third formation, graded by the harness with the serving roles on every node (review round)**:

| check | result |
|---|---|
| `pod_restarts` | **pass** — `0 pod restart(s) within the threshold of 0, 0 OOMKilled` |
| `serving_roles_all_nodes` / `sites_serving_all_nodes` | pass — `2/2 member(s) …; dns agents 3/3, dhcp agents 3/3 registered` / `3/3 nodes each answer the site's DHCP DISCOVER and DNS query` |
| `cluster_formation` / `_unrepaired` · `k3s_restarts` · `gate_predicate` | first try / no repair · `+0/+0/+0` · `3/3 ready, HA, CNPG=3` |
| rollout exit codes | five of five containers the rollout stopped with a running agent: **exit 0** (bind and kea, three nodes); one bind container SIGTERMed 282 ms after start, before its handlers existed: 143 (see §2); the revision-1 key-less pair: 2 (#1062) |

**D. Stuck-agent drill (appliance chart — the shape #1056 was observed on): the bounded wait, proven.** One member's DNS server row set `pending_approval`, its cached bundle wiped, its pod deleted: the agent registered, heartbeated every 30 s and never received a bundle, for 12 min 14 s.

| | observed |
|---|---|
| the wait in `kubectl logs --tail` | `dns_daemon_start_deferred_waiting` re-logged with `waited_s` on the backoff — 1, 2, 4, 8, 16, 32, 60 s within each container's life (the drill ran one commit before the from-loop-start measurement and read 0, 1, 3, 7, 15, 31, 59) — beside `sync_pending_approval_waiting` every 10 s and `heartbeat 200 OK` every 30 s |
| the bound | the liveness probe (tcp :53 — `named` never bound it) restarted the container every ~120 s: containers started 02:57:56, 02:59:56, 03:01:56, 03:03:55, 03:05:56 …; **six restarts in 12 min 14 s**; `Killing … failed liveness probe, will be restarted` |
| how each container ended | `dns_agent_signal_received` → `dns_agent_exiting` ~0.5 s later; `Last State: Terminated, Reason: Completed, Exit Code: 0`, every time (pre-fix: exit 2, `dns_agent_thread_died`) |
| restore (`pending_approval = false` at 03:10:05Z) | `GET …/config 200` 03:10:08.068 → `named_started` :08.091 → `structural_reload_applied` :08.092 → `dns_daemon_launched_after_deferred_start waited_s 10.2` :08.195; `version.bind` answered from 03:10:43; restarts frozen at 6 |
| the control plane's view, the whole 12 minutes | `last_seen_at` advancing every ~60 s, `config_apply_status ok`, `daemon_version 9.20.27`; the API log only `POST …/heartbeat 200 OK`. The heartbeat carried `daemon.status = degraded` throughout and nothing read it — reported separately. |

Physical floor for the wait: the first bundle needs register → `build_config_bundle` → the long-poll response (`DNS_AGENT_LONGPOLL_TIMEOUT` 30 s at most, immediate when the bundle is new); on four fixed members it took 0.71, 0.87, 1.01 and 1.10 s from the deferral — the old code lost that race whenever it exceeded 1.0 s, which is exactly the +1/+1/+2 variance the issue reports.

## Seen alongside, not fixed here: revision-1 pods without an agent key (#1062)

On every formation the DaemonSets go through three revisions. Revision 1 (rendered at the first HelmChart apply) carries **no** `DNS_AGENT_KEY` / `AGENT_GROUP` for bind and no `SPATIUM_AGENT_KEY` / `AGENT_GROUP` for kea; revision 2, 12–20 s later, carries both; revision 3 (once the serving roles land) re-points `CONTROL_PLANE_URL` from the in-cluster Service to the external https URL and adds `SPATIUM_INSECURE_SKIP_TLS_VERIFY=1`. Every revision-1 pod exits 2 within 0.2 s of starting — containerd `exit_status:2` for the bind and kea containers on both clusters — because the entrypoints refuse an empty key (`: "${DNS_AGENT_KEY:?…}"`, `: "${SPATIUM_AGENT_KEY:?…}"`) and busybox `sh` exits 2 on that expansion; the revision-2 rollout then deletes and replaces them. Those exit-2s are the chart's ordering, not the agents', and they are what a reader will find next to this fix's log lines on a fresh install. Filed as #1062.

## Other agents

- **PowerDNS** — same deferral (`pdns_conf_missing_startup_deferred`), same supervisor loop, covered by every change here and unit-tested.
- **Technitium** — never defers on config (starts `dotnet` unconditionally); the only change is that a missing binary now waits for the liveness probe instead of exiting at +1 s. Still loud in its log, and now in its heartbeat.
- **DHCP** — not this shape for the deferred start: `kea-dhcp4`/`kea-dhcp6` are started by the image entrypoint's supervise loops from the shipped idle configs, and the DHCP supervisor watches only its own threads (its pods read 0 restarts on every formation). Its loop has the same thread-death-after-stop order as cause 2, but the container's exit code is the entrypoint's (0 at every rollout observed), so nothing observable changes; left as is.

## Tests

`agent/dns/tests/test_supervisor_deferred_start.py` (16 tests) pins the drivers' deferred / spawned / adopted states and the supervisor loop with its tick scripted: it waits through a deferred start, exits 2 on a real death, exits 0 on a requested stop even when the stop takes the daemon and the threads — including a stop that lands *inside* the daemon check or between the two checks — keeps the boot-time-spawn contract, exits 2 on a thread that dies while not stopping, never checks a driver that manages no daemon, re-logs the wait on the schedule, carries `degraded` in the heartbeat while waiting and clears it on launch without touching a `sync.py` verdict. Each group fails on the commit under it: the loop test exits 2 at tick 1 on the unfixed code; the stop cases exit 2 on the first commit alone; the race cases exit 2 on the second; the visibility cases fail on the third.

- `agent/dns`: 292 passed, 8 skipped, 0 failed (was 276 / 8 / 0) — CI's shape on `python:3.12` (the images and `ci.yml` pin 3.12; the same counts on 3.14).
- Charts: `Charts — Lint & Template` mirrored (`charts-render-check.sh`: helm lint --strict, helm template with every toggle, kubeconform, helm v3.21.4 + kubeconform v0.8.0 in a container): all passed; the rendered `dns-agent` StatefulSet carries both probes.
- ruff on the touched files: clean but for a pre-existing `RUF100` (agent code is not under CI lint).
- gitleaks: no new findings against the base.
- CHANGELOG: `Unreleased` → `Fixed`.

## Tradeoffs and remaining work

- The agent bounds its own wait only through the log and the heartbeat; the probes do the restarting. If a deployment wants a hard cap inside the agent, that is a follow-up knob, not this fix.
- A stop that arrives while the daemon was *never* launched also exits 0 (the loop was waiting); that is the honest reading of a requested stop.
- The heartbeat's `daemon` field is discarded server-side today; reading it is the control plane's follow-up.

## Verification still to come

The QA side carries every open fix of ours, stacked on the next nightly's bytes, through a PostQA walk of both topologies (single node, then the three-node cluster). This fix rides that walk; the `cluster_gate/pod_restarts` result on the stacked build will be posted here, as was done on #1053.

### DIFF
--- a/agent/dns/spatium_dns_agent/drivers/base.py
+++ b/agent/dns/spatium_dns_agent/drivers/base.py
@@ -26,6 +26,13 @@ RRSET_OP_KINDS = frozenset({"create", "update", "delete"})
 
 
 class DriverBase(ABC):
+    #: PID of the daemon this driver spawned or adopted; ``None`` until then.
+    #: Every driver sets it at its spawn / adopt points (``start_daemon`` and
+    #: the system-wide look-up in ``daemon_running``) and never clears it, so
+    #: it is also the "has the daemon been launched at all" fact that
+    #: :meth:`daemon_launched` reports.
+    daemon_pid: int | None = None
+
     def __init__(self, state_dir: Path):
         self.state_dir = state_dir
 
@@ -62,6 +69,21 @@ class DriverBase(ABC):
     @abstractmethod
     def daemon_running(self) -> bool:
         ...
+
+    def daemon_launched(self) -> bool:
+        """True once this driver has spawned or adopted its daemon.
+
+        ``start_daemon`` does not always start one: the BIND9 and PowerDNS
+        drivers return WITHOUT a daemon when no config has been rendered yet
+        (``named_conf_missing_startup_deferred`` /
+        ``pdns_conf_missing_startup_deferred``) and leave the launch to
+        ``swap_and_reload``, which the sync loop reaches once the control
+        plane hands over the first bundle. A daemon that was never launched
+        cannot have died, so the supervisor consults this before it reads
+        ``daemon_running() == False`` as "exit and let the orchestrator
+        restart us" (#1056).
+        """
+        return self.daemon_pid is not None
 
     def daemon_version(self) -> str | None:
         """Version of the DNS daemon binary, e.g. ``"5.0.5"`` / ``"9.20.26"``.
--- a/agent/dns/spatium_dns_agent/supervisor.py
+++ b/agent/dns/spatium_dns_agent/supervisor.py
@@ -28,6 +28,32 @@ from .sync import SyncLoop
 
 log = structlog.get_logger(__name__)
 
+#: What the heartbeat's ``daemon`` field carries while the daemon start is
+#: deferred (no bundle rendered yet, #1056). ``sync.py`` sets its own
+#: degraded verdicts on a failed apply and clears them after the next good
+#: poll; this one is set when the wait begins and cleared by the supervisor
+#: the moment the daemon is up — and only if it is still the supervisor's
+#: own, so a sync verdict set meanwhile is left alone.
+DEFERRED_DAEMON_STATUS: dict[str, str] = {
+    "status": "degraded",
+    "reason": "start deferred, no bundle yet",
+}
+
+
+def wait_log_due(ticks_waiting: int) -> bool:
+    """Whether the waiting state is logged on this tick (1 s each): the first
+    tick, then doubling to 32 s, then every minute. With the crash loop gone
+    this line is nearly the only signal a stuck agent produces, so it has to
+    be in a ``kubectl logs --tail`` an hour in — without a line a second."""
+    return ticks_waiting in (1, 2, 4, 8, 16, 32) or (
+        ticks_waiting >= 60 and ticks_waiting % 60 == 0
+    )
+
+
+def _clear_deferred_status(heartbeat) -> None:
+    if heartbeat.daemon_status.get("reason") == DEFERRED_DAEMON_STATUS["reason"]:
+        heartbeat.daemon_status = {"status": "ok"}
+
 
 def _select_driver(cfg: AgentConfig) -> DriverBase:
     if cfg.driver == "bind9":
@@ -146,6 +172,7 @@ def run(cfg: AgentConfig) -> int:
     # liveness probe (tcp :53) still bounds the wait if no bundle ever comes.
     daemon_managed_drivers = {"bind9", "powerdns", "technitium"}
     waiting_since: float | None = None
+    waiting_ticks = 0
     while not stopping.is_set():
         time.sleep(1.0)
         # A stop requested during the tick (SIGTERM / SIGINT: a DaemonSet
@@ -167,6 +194,8 @@ def run(cfg: AgentConfig) -> int:
                         waited_s=round(time.monotonic() - waiting_since, 1),
                     )
                     waiting_since = None
+                    waiting_ticks = 0
+                    _clear_deferred_status(heartbeat)
             elif driver.daemon_launched():
                 # The stop check above closes the 1 s sleep, not the checks
                 # themselves: ``_sig`` runs between bytecodes, so a SIGTERM
@@ -176,14 +205,19 @@ def run(cfg: AgentConfig) -> int:
                 if not stopping.is_set():
                     log.error("dns_daemon_exited", driver=cfg.driver)
                     return 2
-            elif waiting_since is None:
-                waiting_since = time.monotonic()
-                log.info(
-                    "dns_daemon_start_deferred_waiting",
-                    driver=cfg.driver,
-                    note="start_daemon spawned nothing (no rendered config yet); "
-                    "the sync loop launches the daemon after the first bundle",
-                )
+            else:
+                if waiting_since is None:
+                    waiting_since = time.monotonic()
+                    heartbeat.daemon_status = dict(DEFERRED_DAEMON_STATUS)
+                waiting_ticks += 1
+                if wait_log_due(waiting_ticks):
+                    log.info(
+                        "dns_daemon_start_deferred_waiting",
+                        driver=cfg.driver,
+                        waited_s=round(time.monotonic() - waiting_since, 1),
+                        note="start_daemon spawned nothing (no rendered config yet); "
+                        "the sync loop launches the daemon after the first bundle",
+                    )
         # If any critical thread died (e.g. the sync loop dropped its
         # token after a 401/404 and self-stopped), exit so the container
         # orchestrator restarts us — bootstrap then re-registers from
--- a/agent/dns/spatium_dns_agent/supervisor.py
+++ b/agent/dns/spatium_dns_agent/supervisor.py
@@ -172,7 +172,12 @@ def run(cfg: AgentConfig) -> int:
     # liveness probes (tcp :53 — the appliance chart's, and the umbrella
     # chart's since #1056) still bound the wait if no bundle ever comes.
     daemon_managed_drivers = {"bind9", "powerdns", "technitium"}
-    waiting_since: float | None = None
+    # The deferred wait can only ever begin here — ``daemon_launched()`` never
+    # goes back to False — so it is measured from the loop's start (a breath
+    # after ``start_daemon`` deferred), not from the tick that first noticed
+    # it: the re-log then reads 1, 2, 4, 8 … s, the schedule's own numbers.
+    loop_started = time.monotonic()
+    waiting = False
     waiting_ticks = 0
     while not stopping.is_set():
         time.sleep(1.0)
@@ -188,13 +193,13 @@ def run(cfg: AgentConfig) -> int:
             break
         if cfg.driver in daemon_managed_drivers:
             if driver.daemon_running():
-                if waiting_since is not None:
+                if waiting:
                     log.info(
                         "dns_daemon_launched_after_deferred_start",
                         driver=cfg.driver,
-                        waited_s=round(time.monotonic() - waiting_since, 1),
+                        waited_s=round(time.monotonic() - loop_started, 1),
                     )
-                    waiting_since = None
+                    waiting = False
                     waiting_ticks = 0
                     _clear_deferred_status(heartbeat)
             elif driver.daemon_launched():
@@ -207,15 +212,15 @@ def run(cfg: AgentConfig) -> int:
                     log.error("dns_daemon_exited", driver=cfg.driver)
                     return 2
             else:
-                if waiting_since is None:
-                    waiting_since = time.monotonic()
+                if not waiting:
+                    waiting = True
                     heartbeat.daemon_status = dict(DEFERRED_DAEMON_STATUS)
                 waiting_ticks += 1
                 if wait_log_due(waiting_ticks):
                     log.info(
                         "dns_daemon_start_deferred_waiting",
                         driver=cfg.driver,
-                        waited_s=round(time.monotonic() - waiting_since, 1),
+                        waited_s=round(time.monotonic() - loop_started, 1),
                         note="start_daemon spawned nothing (no rendered config yet); "
                         "the sync loop launches the daemon after the first bundle",
                     )
--- a/agent/dns/spatium_dns_agent/supervisor.py
+++ b/agent/dns/spatium_dns_agent/supervisor.py
@@ -168,8 +168,14 @@ def run(cfg: AgentConfig) -> int:
                     )
                     waiting_since = None
             elif driver.daemon_launched():
-                log.error("dns_daemon_exited", driver=cfg.driver)
-                return 2
+                # The stop check above closes the 1 s sleep, not the checks
+                # themselves: ``_sig`` runs between bytecodes, so a SIGTERM
+                # that lands inside ``daemon_running()`` (a rollout that took
+                # named first) arrives here with the daemon gone and the stop
+                # already requested. A stop is a stop, whenever it lands.
+                if not stopping.is_set():
+                    log.error("dns_daemon_exited", driver=cfg.driver)
+                    return 2
             elif waiting_since is None:
                 waiting_since = time.monotonic()
                 log.info(
@@ -183,7 +189,7 @@ def run(cfg: AgentConfig) -> int:
         # orchestrator restarts us — bootstrap then re-registers from
         # PSK with a fresh empty token cache.
         dead = [t.name for t in threads if not t.is_alive()]
-        if dead:
+        if dead and not stopping.is_set():
             log.error("dns_agent_thread_died", threads=dead)
             return 2
 
--- a/agent/dns/spatium_dns_agent/supervisor.py
+++ b/agent/dns/spatium_dns_agent/supervisor.py
@@ -168,8 +168,9 @@ def run(cfg: AgentConfig) -> int:
     # container until the bundle happened to arrive inside the 1 s window
     # (#1056: ``dns_daemon_exited`` 1.0 s after the deferral, Last State
     # exit 2, restart count +2 on a member). So: wait while nothing has
-    # been launched, exit only when a launched daemon is gone. The chart's
-    # liveness probe (tcp :53) still bounds the wait if no bundle ever comes.
+    # been launched, exit only when a launched daemon is gone. The charts'
+    # liveness probes (tcp :53 — the appliance chart's, and the umbrella
+    # chart's since #1056) still bound the wait if no bundle ever comes.
     daemon_managed_drivers = {"bind9", "powerdns", "technitium"}
     waiting_since: float | None = None
     waiting_ticks = 0
--- a/agent/dns/spatium_dns_agent/supervisor.py
+++ b/agent/dns/spatium_dns_agent/supervisor.py
@@ -129,12 +129,45 @@ def run(cfg: AgentConfig) -> int:
     # process (BIND9, PowerDNS, Technitium). If the daemon dies, we exit
     # non-zero so the orchestrator restarts the container — agent + daemon
     # are bound lifecycle-wise.
+    #
+    # "Dies" presumes it was launched. ``start_daemon()`` above returns
+    # WITHOUT a daemon when no config has been rendered yet (BIND9
+    # ``named_conf_missing_startup_deferred``, PowerDNS
+    # ``pdns_conf_missing_startup_deferred``) and the sync loop launches it
+    # from ``swap_and_reload`` once the first bundle lands. The first bind
+    # pod on a freshly joined cluster member always boots that way — its
+    # state dir is empty and the control plane has not built its bundle
+    # yet — and reading that deferred start as a death made this loop
+    # return 2 one tick after boot: kubelet then back-off-restarted the
+    # container until the bundle happened to arrive inside the 1 s window
+    # (#1056: ``dns_daemon_exited`` 1.0 s after the deferral, Last State
+    # exit 2, restart count +2 on a member). So: wait while nothing has
+    # been launched, exit only when a launched daemon is gone. The chart's
+    # liveness probe (tcp :53) still bounds the wait if no bundle ever comes.
     daemon_managed_drivers = {"bind9", "powerdns", "technitium"}
+    waiting_since: float | None = None
     while not stopping.is_set():
         time.sleep(1.0)
-        if cfg.driver in daemon_managed_drivers and not driver.daemon_running():
-            log.error("dns_daemon_exited", driver=cfg.driver)
-            return 2
+        if cfg.driver in daemon_managed_drivers:
+            if driver.daemon_running():
+                if waiting_since is not None:
+                    log.info(
+                        "dns_daemon_launched_after_deferred_start",
+                        driver=cfg.driver,
+                        waited_s=round(time.monotonic() - waiting_since, 1),
+                    )
+                    waiting_since = None
+            elif driver.daemon_launched():
+                log.error("dns_daemon_exited", driver=cfg.driver)
+                return 2
+            elif waiting_since is None:
+                waiting_since = time.monotonic()
+                log.info(
+                    "dns_daemon_start_deferred_waiting",
+                    driver=cfg.driver,
+                    note="start_daemon spawned nothing (no rendered config yet); "
+                    "the sync loop launches the daemon after the first bundle",
+                )
         # If any critical thread died (e.g. the sync loop dropped its
         # token after a 401/404 and self-stopped), exit so the container
         # orchestrator restarts us — bootstrap then re-registers from
--- a/agent/dns/spatium_dns_agent/supervisor.py
+++ b/agent/dns/spatium_dns_agent/supervisor.py
@@ -148,6 +148,16 @@ def run(cfg: AgentConfig) -> int:
     waiting_since: float | None = None
     while not stopping.is_set():
         time.sleep(1.0)
+        # A stop requested during the tick (SIGTERM / SIGINT: a DaemonSet
+        # rollout, a scale-down, an operator) has already stopped every
+        # worker thread through ``_sig`` and may have taken the daemon with
+        # it. Whatever the checks below would find dead now died because we
+        # were told to stop — take the designed exit, not the crash exits,
+        # so the container's last state reads 0 rather than "thread died"
+        # / "daemon exited" (#1056: every rollout ended the old container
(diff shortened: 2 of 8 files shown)


agent/dns/spatium_dns_agent/supervisor.py
```diff
@@ -129,12 +129,55 @@ def _sig(_signum, _frame):  # noqa: ANN001
...
+    # from ``swap_and_reload`` once the first bundle lands. The first bind
+    # pod on a freshly joined cluster member always boots that way — its
+    # state dir is empty and the control plane has not built its bundle
+    # yet — and reading that deferred start as a death made this loop
+    # return 2 one tick after boot: kubelet then back-off-restarted the
+    # container until the bundle happened to arrive inside the 1 s window
+    # (#1056: ``dns_daemon_exited`` 1.0 s after the deferral, Last State
+    # exit 2, restart count +2 on a member). So: wait while nothing has
+    # been launched, exit only when a launched daemon is gone. The chart's
+    # liveness probe (tcp :53) still bounds the wait if no bundle ever comes.
```
mzac: This is the claim I'd most want to adjust before merge: the agent gives up its own bound and hands it to the probe, and on one of the two supported k8s shapes that probe does not exist.

- `charts/spatiumddi-appliance/templates/{dns-bind9,dns-powerdns,dns-technitium}.yaml` each carry a readinessProbe **and** `livenessProbe: tcpSocket :53` (30 s initial, 30 s period). The sentence is exactly right there — and that is the shape #1056 was observed on.
- `charts/spatiumddi/templates/dns-agent.yaml:137` carries a readinessProbe only (`tcpSocket :53`, 5 s/10 s). No livenessProbe anywhere in that file, and a readiness probe never restarts a pod.

So on the umbrella chart the wait is unbounded rather than bounded. An agent whose first bundle never arrives — wrong `AGENT_GROUP`, a revoked PSK, a control plane reachable for `/heartbeat` but not producing a bundle — now sits forever: NotReady, registered, heartbeating, `last_seen_at` fresh. Before this it exited 2 and showed CrashLoopBackOff. That path is supported and `agent-e2e.yml` exercises it with `dnsAgents.enabled=true`.

Either add the matching livenessProbe to the umbrella chart here (non-negotiable #12 asks the manifests to track each other) or narrow this sentence and the PR body to say *the appliance chart's* liveness probe.

agent/dns/spatium_dns_agent/supervisor.py
```diff
@@ -129,12 +129,55 @@ def _sig(_signum, _frame):  # noqa: ANN001
...
-            return 2
+        # A stop requested during the tick (SIGTERM / SIGINT: a DaemonSet
+        # rollout, a scale-down, an operator) has already stopped every
+        # worker thread through ``_sig`` and may have taken the daemon with
+        # it. Whatever the checks below would find dead now died because we
+        # were told to stop — take the designed exit, not the crash exits,
+        # so the container's last state reads 0 rather than "thread died"
+        # / "daemon exited" (#1056: every rollout ended the old container
+        # with exit 2 and a dns_agent_thread_died in its log).
+        if stopping.is_set():
```
mzac, line 192: This closes the stop that lands during the 1 s sleep, but not one that lands *after* this check. Python runs signal handlers between bytecodes in the main thread, so `_sig` can fire anywhere between here and the `if dead:` block — stopping every worker thread, and a rollout can take the daemon in the same instant.

Both exits stay reachable. Reproduced against this head using your own `supervised` fixture, firing the stop from inside `daemon_running()`:

```
daemon-check race -> rc: 2 | events: ['dns_agent_signal_received', 'dns_daemon_exited']
thread-check race -> rc: 2 | events: ['dns_agent_signal_received', 'dns_agent_thread_died']
```

The window went from ~1 s to the duration of the two checks, so this is rare now rather than every rollout — but by the PR's own framing it is the same defect, and guarding the two `return 2` sites with `not stopping.is_set()` closes it in two lines with no new state. Worth noting `test_a_stop_is_exit_0_even_when_it_takes_the_daemon_and_the_threads` cannot catch it: its script fires the stop inside the patched sleep, so it only ever exercises the path this `break` already handles.

agent/dns/spatium_dns_agent/supervisor.py
```diff
@@ -129,12 +129,55 @@ def _sig(_signum, _frame):  # noqa: ANN001
...
+                    log.info(
+                        "dns_daemon_launched_after_deferred_start",
+                        driver=cfg.driver,
+                        waited_s=round(time.monotonic() - waiting_since, 1),
+                    )
+                    waiting_since = None
+            elif driver.daemon_launched():
+                log.error("dns_daemon_exited", driver=cfg.driver)
+                return 2
+            elif waiting_since is None:
```
mzac: `elif waiting_since is None` means this fires exactly once, at t=1 s, and never again. With the crash loop gone that is very nearly the only signal a stuck agent produces.

The other two channels are both empty. The supervisor never touches `heartbeat.daemon_status` — it is written in exactly two places, both in `sync.py`'s config-apply path (`:532` degraded on failure, `:443` back to ok) — so an agent that never received a bundle leaves it `{}`. And the `daemon` field the agent does send is discarded by the backend (see the review body).

So on the umbrella chart or on Compose, a permanently-waiting agent is: registered, heartbeating on schedule, `last_seen_at` fresh, one log line at the very start, nothing in the control plane. Someone running `kubectl logs --tail` an hour later sees no reason at all.

Either would fix it, and both are small:
- re-log on a backoff (every Nth tick) so the state is present in a tail rather than only in the first second;
- set `heartbeat.daemon_status = {"status": "degraded", "reason": "start deferred, no bundle yet"}` while waiting, clearing it alongside `waiting_since`. That also gives the unread `daemon` field its first consumer, which seems like the right place for it to start.

agent/dns/spatium_dns_agent/drivers/base.py
```diff
@@ -26,6 +26,13 @@
 
 
 class DriverBase(ABC):
+    #: PID of the daemon this driver spawned or adopted; ``None`` until then.
+    #: Every driver sets it at its spawn / adopt points (``start_daemon`` and
+    #: the system-wide look-up in ``daemon_running``) and never clears it, so
+    #: it is also the "has the daemon been launched at all" fact that
+    #: :meth:`daemon_launched` reports.
+    daemon_pid: int | None = None
```
mzac, line 34: Minor, and only a description/tidiness point: the PR body says the declaration "moves to the base class", but no driver file is in this diff and all three still declare it — `bind9.py:713`, `powerdns.py:290`, `technitium.py:618`. So this is a fourth copy, not a move.

No runtime effect (all four default to `None`, so an identical shadow is a no-op), but the base class is now the documented home of a fact three subclasses re-declare. Either drop the three subclass lines here or adjust the description.

The substance of the comment is right, though, and I checked it: every `daemon_pid` assignment across the three drivers is a real pid and there is no `= None` after the defaults, so `daemon_launched()` cannot regress to False once armed.

mzac (commented):
Reviewed at `3da5b5d` in a clean worktree. Both diagnoses are correct and specific, and the test file is the strongest part of this PR.

Two things I checked rather than took on trust, both of which came back in your favour:

- **The suite reproduces exactly** — 286 passed, 8 skipped, py3.14 from a clean export.
- **The negative control holds.** With `base.py` and `supervisor.py` reverted to `origin/main` and the new test file kept, **8 of the 10 fail**. They test the fix, not the mock.
- The load-bearing premise is true: `daemon_pid` is assigned only real pids across bind9 / powerdns / technitium, there is no `daemon_pid = None` anywhere after the class defaults, so `daemon_launched()` cannot regress to False after a launch and a genuine post-launch death still exits 2.

@amoona6 — five points below, four inline. My read is that #1 (line 146) and #3 (line 173) are the pair worth doing here, because together they turn a loud failure into a silent one on a supported install path; #2 is two lines. None of them is an argument against the change — the crash loop was worse than the wait.

Two that have no line in this diff to hang on:

### No CHANGELOG entry

The diff is three files and none is `CHANGELOG.md`. The behaviour is operator-visible twice over: restart count 0 instead of +2 on a freshly joined member, and a rollout ending the old container 0 rather than 2 with a `dns_agent_thread_died` in its log. That second one especially is what someone greps the changelog for after noticing it changed between releases. `## Unreleased` → `### Fixed`, same as #1060.

### The heartbeat's `daemon` field is written and never read

Context for the inline comment at line 173. `heartbeat.py:106` puts `"daemon": self.daemon_status` on the wire and `backend/app/api/v1/dns/agents.py:108` declares `daemon: dict[str, Any] = {}` — and no handler code reads `body.daemon` (only `body.daemon_version`, at :452). Pre-existing, and the class #882 named, but it is why the new waiting state has nowhere to surface.

---

Mechanically: 24/24 green, `MERGEABLE`, `BEHIND` main.

agent/dns/spatium_dns_agent/supervisor.py
amoona6: Confirmed — the sentence held for the appliance chart only. Rather than narrow it, 9308efb7 gives the umbrella chart's dns-agent StatefulSet (all three flavours) the appliance chart's exact liveness probe (tcpSocket :53, 30 s initial, 30 s period), so the manifests track each other and a never-arriving bundle is a bounded restart on both shapes. Charts — Lint & Template mirrored (helm lint --strict, template with every toggle, kubeconform): passed; the rendered StatefulSet carries both probes. Proven live on the appliance chart with a stuck-agent drill (details in the summary comment): six probe restarts in 12 min 14 s, each container ending Completed / exit 0, recovery within 3 s of the bundle arriving.

agent/dns/spatium_dns_agent/supervisor.py
amoona6: Confirmed. 3e91d746 guards both crash exits with `not stopping.is_set()`; the residual is the few bytecodes between that test and the return. Two tests fire the stop from inside `daemon_running()` (daemon dead / daemon up): rc 2 on the previous head, 0 now.

agent/dns/spatium_dns_agent/supervisor.py
amoona6: Confirmed. ee9cb60b re-logs the wait on a backoff (ticks 1, 2, 4, 8, 16, 32, then every 60 s, each with `waited_s`) and sets `heartbeat.daemon_status = {status: degraded, reason: start deferred, no bundle yet}` while waiting, cleared to ok on launch only if it is still the supervisor's own verdict. Four tests; three fail on the commit beneath. In the drill a container's `--tail` reads the backoff beside the sync loop's own waiting line. As you said, the field is not observable server-side — the drill's 12 minutes show `last_seen_at` advancing and `config_apply_status ok` throughout — so the control-plane gap is filed separately (linked in the summary comment).

agent/dns/spatium_dns_agent/drivers/base.py
amoona6: Agreed — a fourth copy, not a move. 142595ab drops the three subclass declarations; the description is corrected.

Issue #219, opened 2026-05-17
[worker] `refresh_blocklist_feed` declares `max_retries=3` but never retries


mzac:
**Component**: `worker`  
**Severity**: `high`  
**Location**: `backend/app/tasks/dns.py:114`  

---

Declares `bind=True, max_retries=3` but the task body never calls `self.retry()` and has no `autoretry_for=(...)` — every exception is swallowed internally and written to the DB row as `last_sync_status='error'`. The `max_retries=3` is a no-op. Transient HTTP failures require the operator to manually re-fire from the UI. Either add `autoretry_for=(httpx.HTTPError, socket.gaierror)` with a backoff, or drop `max_retries=3` to stop signaling retry-capability that doesn't exist.

---
_Filed automatically from the `mzac-bug-check` codebase audit pass._

mzac:
Fixed in PR #268 (merged 2026-05-18 as commit `80b7ba5`). See the PR description for the per-issue change summary.

Issue #216, opened 2026-05-17
[api] `SECRET_KEY` default `change-me-to-a-random-32-char-string`


mzac:
**Component**: `api`  
**Severity**: `low`  
**Location**: `backend/app/config.py:16`  

---

`secret_key: str = "change-me-to-a-random-32-char-string"` default. `.env.example` should force operators to set one (`openssl rand -hex 32`). Add a `model_validator` that rejects the literal default in non-dev mode so a misconfigured prod deploy fails to boot rather than silently signing JWTs with a public key.

---
_Filed automatically from the `mzac-bug-check` codebase audit pass._

PR #1063, opened 2026-09-11
perf(orchestrator): count DNS answers by rcode, attribute DHCP ACKs to the exchange they answer, and read both ledgers in the report (#1057)

### LINKED ISSUE #1057: perf orchestrator: a DNS response with any rcode but NOERROR/NXDOMAIN is counted as nothing, and a DORA/renew ACK that arrives after the timeout is a timeout — the handshake and answer figures cannot be trusted
Two accounting holes in `perf/generators/orchestrator/device_fleet.py`, seen on two consecutive cluster load runs of the nightly QA harness (2026-09-10 and 2026-09-11, `nightly-2026.09.09` and `nightly-2026.09.10` plus PR #1053), where the appliance's own counters say the opposite of the generator's.

**1. DNS.** `_dns_query` (`device_fleet.py:776-794`) increments `dns_ok` only for `NOERROR`/`NXDOMAIN` and `dns_timeout` only on an exception. Any other rcode — REFUSED, SERVFAIL, FORMERR — is counted as neither, so the run's `dns_ok / dns_sent` reads as a loss that is not a loss and the report cannot say what happened. On 2026-09-11 the generator recorded `dns_sent ≈ 606k, dns_ok 0, dns_timeout 28`; BIND's statistics channel on the target recorded `QUERY 606,811 → REFUSED 606,810, NOERROR 1` (the generator's source address matched no view). The previous night: `dns_sent 606,185 / dns_ok 0 / dns_timeout 46` while the generator's own `dns_resolve` latency histogram held 606,139 samples — every query got a response the counters did not name. The fix is one dictionary: count responses per rcode (`dns_rcode_<name>`) and define `dns_ok` explicitly; an out-of-view or ACL problem then reads as "refused", not as vanished queries.

**2. DHCP.** `_on_ack` (`device_fleet.py:503-540`) counts `dora_ack` only when the device is still `DISCOVERING`; `dora_timeout` is scheduled at `DORA_TIMEOUT_S` (`:433`, `:474`). An ACK that lands after the timeout moved the device on is counted as a timeout AND discarded, so under any latency near the timeout the handshake figure undercounts what the server actually delivered. On 2026-09-11 the generator reported `handshake 94.01 % (acks 9,997, timeouts 637)` while kea's log on the appliance showed `14,074 REQUESTs received, 14,074 ACKed, 0 NAK, 0 dropped`. The previous night: `dora_sent 19,520 / dora_ack 8,599 / timeouts 695` — about 10,200 DORAs neither acked nor timed out in the counters — and `renew_ack 0` against `renew_sent 4,878` beside `rebind_ack 4,601`. Suggested: count a late ACK as `dora_ack_late` (and `renew_ack` regardless of the state the reply finds), and report the handshake figure with and without late arrivals; keep the timeout as the SLO input if that is the intent, but never lose the ACK.

Both figures feed our Tier 0 verdicts on the cluster load tier, which is why they are here rather than in a note: with these holes the tier cannot distinguish a lossy appliance from a lossy count. The appliance-side counters we cross-check against (kea's log tally, BIND's statistics channel) are what surfaced it.

amoona6:
Fixes #1057 — the device-fleet orchestrator counted DNS answers only when they were NOERROR/NXDOMAIN and DHCP ACKs only while the device's state still expected them, so its handshake and DNS figures could not be reconciled with kea's and BIND's own counters for the same packets.

Root-caused and validated live on a nested single-node QA appliance driven by the same `gate` manifest from the same load-generator host, once with the unfixed `perf/` tree at `ca1b6bfd` and once with this branch, with kea's `statistic-get-all` (`pkt4-*`), kea's `dhcp_metric_sample` buckets and BIND's JSON statistics channel read inside the appliance before and after each run; then again under a 9 s one-way network delay on DHCP alone, which is the only way to make the late-ACK path fire deterministically. Unit tests drive the orchestrator's FSM without a socket for the same paths.

## 1. `_dns_query` counted every non-NOERROR/NXDOMAIN answer as nothing

`perf/generators/orchestrator/device_fleet.py` (at `ca1b6bfd`, lines 785-794): `dns_ok += 1` only for rcode NOERROR/NXDOMAIN, `dns_timeout += 1` in a bare `except Exception`. An answer with any other rcode incremented `dns_sent` and the `dns_resolve` histogram and nothing else; `dns_timeout` also swallowed every non-timeout exception.

Live, same appliance, same manifest, ~606k queries, every one answered REFUSED by BIND (the generator's UDP source matches no view — a fact about the vantage, which the numbers should say, and did not):

| | unfixed | fixed | BIND, Δ over the run (unfixed / fixed) |
|---|---|---|---|
| `dns_sent` | 606,196 | 606,237 | `opcodes.QUERY` 606,549 / 606,800 — the excess is the propagation probe's own `dig` polls |
| `dns_answered` | — | **606,237** | `nsstats.Response` 606,549 / 606,800 |
| `dns_ok` / `dns_timeout` | 0 / 0 | 0 / 0 | `NOERROR` 0 / 0 |
| `dns_rcode_REFUSED` | (not counted) | **606,237** | `REFUSED` 606,549 / 606,800 |
| answers counted as nothing | **606,196** | 0 | |
| report row b6a (REFUSED = 0, structural) | NO_DATA | **FAIL** `606237 (dnsperf 0, orchestrator 606237)` | |

The unfixed run's histogram held all 606,196 answers (p50 25.8 ms) while its counters said "ok 0, timeouts 0"; the run this issue was filed from said "ok 0, timeouts 46" for the same reason.

**Fix (commit 1):** every answer is counted under `dns_rcode_<NAME>` (counter snapshot, periodic stats, final summary); `dns_answered` counts every response; `dns_ok` keeps its meaning (NOERROR + NXDOMAIN); `dns.exception.Timeout` → `dns_timeout`, any other exception → `dns_error` (logged once per type). The final summary carries a `dns` block computed by the new `spddi_perf.generator_tallies` fold.

## 2. `_on_ack` decided what an ACK meant from the device's state; the DORA deadline timers were per packet

`device_fleet.py` (at `ca1b6bfd`, lines 505-567, 593-603, 433, 474). `_on_ack` branched on `prev_state`: DISCOVERING → DORA ack, RENEWING/REBINDING → renew/rebind ack, anything else → nothing. After `MAX_DORA_RETRIES` the device is OFFLINE with `timeout += 1`, so the ACK that answers its last REQUEST matched no branch: the timeout stood and the lease kea had just allocated was discarded. Each DISCOVER and each REQUEST scheduled its own `dora_timeout` against the device, so a DISCOVER's deadline could retransmit over a live REQUEST. And an ACK from a previous round landing on a device that had already re-arrived was counted as the new round's DORA, with a latency measured from the wrong send.

Live, under a 9 s one-way delay on UDP/67 into the appliance (DNS untouched: `dig` still answered in 20 ms), 400-device manifest, 12 minutes, unfixed then fixed:

| | unfixed | fixed | kea, Δ over the run (unfixed / fixed) |
|---|---|---|---|
| DISCOVERs sent | 10,566 | 5,679 | `discover-received` 10,357 / 5,650 |
| OFFERs received / ignored (device already leased) | — | 5,650 / 1,685 | `offer-sent` 10,357 / 5,650 |
| REQUESTs sent | — | 3,965 | `request-received` 3,385 / 3,962 |
| `dora_ack` (before the device gave up) | 1,341 | 1,160 (all over budget) | `ack-sent` 3,385 / 3,962 |
| `dora_ack_late` (after it gave up) | — | **241** | |
| `ack_duplicate` (ACKs to REQUESTs the device had superseded) | — | 2,561 | |
| `timeout` | 1,460 | 241 | `receive-drop` 0 / 0 |
| handshake strict / with late | 47.9 % / — | 82.8 % / **100 %** | kea served 100 % / 100 % |
| ACKs kea sent vs ACKs the generator accounted for | 3,385 vs 1,341 — **2,044 dropped or misattributed** | 3,962 vs 1,160 + 241 + 2,561 = **3,962, exact** | |
| DORA latency p50 | "535 ms" under a 9 s delay — the ACK was attributed to a later round's send | 9,011 ms (`dora_ack_late` p50 9,011 ms) | |

With every ACK at least 18 s away, the unfixed device gives up at ~13 s (four per-packet deadlines) before any ACK arrives; the ACKs that then land are dropped (device OFFLINE) or, when the device has re-arrived, counted as the new round's DORA. The fixed tree attributes each ACK to the REQUEST it answers: 1,160 answered a round still in progress, 241 answered after the device had given up (every one of its 241 timeouts turned out to be a slow ACK), 2,561 answered REQUESTs the device had superseded (a second OFFER accepted while the first REQUEST's ACK was in the pipe), and 1,685 OFFERs arrived after the device already held a lease (5,650 − 3,965, exact).

The unstressed full run tells the other half of the story. Fixed tree, 4,000 devices, 25 minutes:

| | fixed generator | kea, Δ over the run |
|---|---|---|
| DISCOVERs sent / OFFERs received | 5,336 / 5,007 | `discover-received` 5,007 / `offer-sent` 5,007 |
| REQUESTs sent: selecting / renew / rebind | 5,007 / 2,791 / 2,676 | `request-received` 9,922 |
| `dora_ack` / `renew_ack` / `rebind_ack` | 4,731 / 0 / 2,497 | `ack-sent` 9,922 |
| `timeout` / lapses / late / over budget | 0 / 179 / 0 / 0 | `receive-drop` 0 |

The OFFER leg proves the return path lost nothing (5,007 = 5,007), so the 552 REQUESTs kea never received (10,474 − 9,922) are inbound loss: 276 selecting (5,007 − 4,731), 179 rebind (exactly the 179 lapses), 97 renew — and kea's 9,922 ACKs are 4,731 + 2,497 received at the relay address plus 2,694 renew ACKs (2,791 − 97) unicast to `ciaddr:68`, which a relay-topology load generator never owns. Every one of the 605 retransmitted DISCOVERs (5,336 − 4,731) is an inbound loss, not a slow reply (`dora_ack_over_budget 0`, DORA p99 78 ms). None of this was visible before: the unfixed run reported "handshake 100 %" over the same loss, and the cluster run this issue came from reported "94 %" with kea serving 100 % of what reached it. Where the inbound loss happens (before kea's `pkt4-received`, with kea's own drop counter at 0 and the socket buffer at 0-25) is a separate investigation the new counters now make possible on every run.

**Fix (commit 2):** an `ExchangeLedger` keyed by xid. A reply is attributed to the exchange it answers, with that exchange's own send time, and a verdict: `duplicate` (already closed), `over_budget` (its own timer had fired — the device retried past it but had not given up) or `late` (the device had given up on the round and counted a timeout / lapse).

- `dora_ack` keeps its pre-fix meaning (ACKed before the device gave up, retries included) so existing consumers read the same figure.
- `dora_ack_late`, `renew_ack_late`, `rebind_ack_late` count what used to be dropped; `dora_ack_over_budget`, `dora_offer`, `offer_ignored`, `request_sent`, `ack_duplicate`, `ack_unmatched` make the tally reconcilable with kea's `pkt4-*` counters, as above.
- A late DORA or rebind ACK takes the lease it carries (the server holds it) unless a newer round already did; a late ACK for a device that departed moves nothing.
- Timers are scoped to their exchange; T1 carries a token so a lease re-ACKed twice (a late renew ACK, then the rebind's) renews once; late DORA latencies go to their own histogram (`dhcp_dora_ack_late`) so the strict one keeps its meaning.
- The final summary carries a `handshake` block at three strictnesses: within budget / strict / with late.

Deliberate behaviour change, stated for review: the retry budget is now per exchange — a REQUEST gets its own `DORA_TIMEOUT_S` instead of inheriting its DISCOVER's remaining deadline — so a device facing a slow OFFER *and* a slow ACK can take up to about twice as long to give up as before. `MAX_DORA_RETRIES` still bounds the attempts.

## 3. The report ignored the orchestrator's rcodes and summed a cumulative counter

`perf/harness/spddi_perf/collect.py` (at `ca1b6bfd`): b6 (SERVFAIL rate) and b6a (REFUSED = 0, a structural invariant per §8.3.1) read dnsperf only (lines 678-695), so a run whose orchestrator stream was entirely refused reported NO_DATA on both. b7 summed `dns_timeout` over every periodic window (line 700), but the field is cumulative per shard, so the printed figure multiplied every earlier window's timeouts into every later row.

**Fix (commit 3):** b6/b6a fold the orchestrator's `dns_rcode_*` in (b6a now FAILS structurally on a refused stream — the run measured nothing, whether the name escaped the §4.9 validator or the generator queried from outside the served view; live: overall `FAIL — structural invariant failed: b6a`); b7 reads the final summary (or the last window per shard) and counts `dns_error` beside the timeouts; `slo_results.json` gains a `generator` block; both report renderers print a "Generator accounting (orchestrator)" section.

## What the gate reports until the vantage is fixed

b6a moves from NO_DATA to a structural FAIL on any orchestrator run whose stream BIND refused. On the QA appliance this change was proven on, that is every run today: the generator's UDP source sits outside the client ACL of every view the appliance serves, so the DNS half of those runs measured nothing — and the report now says so instead of saying nothing. The vantage fix is queued on the QA side (ddi-pg #216: an in-view source address for the generator's DNS stream). Until it lands, the perf gate on that appliance reads `overall FAIL — structural invariant failed: b6a` by construction, and the QA harness turns its DNS rows into an infrastructure skip that names the vantage (ddi-pg #209, merged) and will treat the SLO row the same way (ddi-pg #219). Once the vantage is in view, b6a is back to measuring what it was written for — an out-of-zone leak — and the first cluster re-measurement after the harness pin bump is expected green on that row.

Both trees ran from the same host against the same appliance with the same manifest and bound groups (the fixed full run with `--skip-seed`, because the seeder cannot run twice against one appliance in relay topology — it creates a new IP space per run_id and POSTs the same scopes into the shared DHCP group, `409 … already covers 10.100.0.0/22`; noted here as a fact about the seeder, not changed). Appliance-side counters bracket each run; the load host's own DHCP/DNS probes counted 0 in every window. The identities above are exact packet counts; the one derived term is the renew split of the full run (2,694 = kea's ACKs − the ACKs received at the relay address, bounded by the 2,791 renews sent), and the DNS excess (BIND's QUERY − `dns_sent`) is the propagation probe's own polls.

| figure | unfixed | fixed |
|---|---|---|
| DNS answers accounted for, full run | 0 of 606,196 | 606,237 of 606,237 (REFUSED, per BIND) |
| report b6a on a refused stream | NO_DATA | FAIL (structural) |
| ACKs accounted for under a 9 s delay | 1,341 of kea's 3,385 | 3,962 of kea's 3,962 |
| handshake under a 9 s delay | 47.9 % | 82.8 % strict, 100 % with late |
| DORA latency under a 9 s delay | p50 "535 ms" | p50 9,011 ms |
| inbound DHCP loss, full run | invisible | 329 DISCOVERs, 552 REQUESTs (of 5,336 / 10,474) |

## Tests

- `python -m pytest perf -q` in the CI job's environment: **69 passed, 0 skipped** (41 at base). The job now installs `pytest pyyaml dnspython hdrhistogram` — the orchestrator's and report's own imports — because with pytest alone the two behavioural test files import-skipped (56 passed, 2 skipped) and a reverted fix stayed green (review round 1; the same reversion fails 12 tests once they run). hdrhistogram ships a wheel for x86_64 only, so the Makefile mirror takes it as a wheel when one exists and the single `.hdr` assertion is gated on the backend.
- New: `perf/generators/orchestrator/tests/test_accounting.py` (the ledger and the tally, pure), `test_device_fleet_accounting.py` (the orchestrator FSM driven without a socket: DISCOVER deadline over a live REQUEST, strict / over-budget / late ACKs, renew escalation then rebind, lapse revival vs departure, NAK restart, DNS by rcode, the summary blocks), `perf/harness/tests/test_generator_tallies.py`, `test_collect_generator_rows.py`.
- `perf/` is outside the Backend Lint scope by `ci.yml`'s own note; the added files are ruff-clean and the two touched files carry only their pre-existing findings. `py_compile` of every perf file passes.

## Tradeoffs and remaining work

- Existing field names keep their meaning — add, never rename — so a consumer that folds `dora_ack` / `dns_ok` / `dns_timeout` today reads the same numbers; the new fields are what let it reconcile them.
- The RENEWING leg is unobservable in relay topology by construction: the renew REQUEST carries `giaddr=0, ciaddr=leased IP` (`dhcp_packet.build_request_renew`), so kea unicasts the ACK to `ciaddr:68`. This PR counts renew ACKs by xid whatever state finds them, so a topology that can see them counts them — it does not change the packet.
- Late ACK latencies are kept in their own histogram rather than folded into `dora_ack`'s p99, so the criterion-(b) rows keep their meaning.
- The seeder's re-runnability against one appliance, and the inbound DHCP loss the new counters expose, are follow-ups, not part of this change.

## Verification still to come

The QA fleet's rigs clone this harness at a pinned ref rather than at the build under test, so the nightly PostQA walk cannot carry a `perf/` change; the fleet-side proof is the pin bump once this merges, after which the cluster load tier is re-measured with the reconciled counters and the result posted here.

### DIFF
--- a/perf/generators/orchestrator/device_fleet.py
+++ b/perf/generators/orchestrator/device_fleet.py
@@ -66,8 +66,11 @@ import dhcp_packet as dp  # noqa: E402
 from lifecycle_log import LatencyAccumulator, LifecycleLog  # noqa: E402
 from relay_sockets import open_relay_sockets  # noqa: E402
 
-from accounting import DnsTally, rcode_name  # noqa: E402
-from spddi_perf.generator_tallies import dns_summary  # noqa: E402
+from accounting import (  # noqa: E402
+    DORA_KINDS, KIND_DISCOVER, KIND_REBIND, KIND_RENEW, KIND_SELECT, Closed, DnsTally,
+    Exchange, ExchangeLedger, rcode_name,
+)
+from spddi_perf.generator_tallies import dns_summary, handshake_summary  # noqa: E402
 
 # dnspython is an off-box runtime dep (perf/requirements.txt: dnspython>=2.6); resolve
 # it once at import so the per-query hot path doesn't re-run the import machinery. When
@@ -135,6 +138,14 @@ class Device:
     dora_retries: int = 0
     # propagation-probe membership
     probe: bool = False
+    # #1057 — exchange-correlated accounting: the DORA / renewal round this
+    # device is on (retransmits keep it, a fresh arrival or a NAK bumps it), the
+    # token of the live T1 timer (a re-ACKed lease invalidates the old one), and
+    # whether a LEFT device got there by lapsing (a late rebind ACK revives it)
+    # or by departing (it does not).
+    episode: int = 0
+    t1_token: int = 0
+    lapsed: bool = False
 
 
 @dataclass
@@ -170,8 +181,18 @@ class Counters:
     ddns_first_publish: int = 0
     ddns_renew_writes: int = 0   # MUST stay 0 (H3) — incremented only if a renewal IP-changes
     renew_ip_changed: int = 0    # named correctness FAIL signal
-    # #1057 — every DNS answer counted. The fields above keep their meaning
-    # (add, never rename: ddi-pg's load tier folds them by name).
+    # #1057 — exchange-correlated accounting. The fields above keep their
+    # meaning (add, never rename: ddi-pg's load tier folds them by name); these
+    # make the generator's tally reconcilable with kea's and BIND's own counters.
+    dora_offer: int = 0            # OFFERs received, whatever the device was doing
+    offer_ignored: int = 0         # OFFERs that found the device not DISCOVERING
+    request_sent: int = 0          # SELECTING REQUESTs sent (one per accepted OFFER)
+    dora_ack_over_budget: int = 0  # of dora_ack: its exchange's own timer had fired
+    dora_ack_late: int = 0         # DORA ACKs after the device gave up (a `timeout` that was slow)
+    renew_ack_late: int = 0        # renew ACKs after the renew timer escalated / the device left
+    rebind_ack_late: int = 0       # rebind ACKs after the device lapsed (a `lapses` that was slow)
+    ack_duplicate: int = 0         # a second ACK for an exchange already closed
+    ack_unmatched: int = 0         # an ACK whose xid matches no exchange the ledger knows
     dns_answered: int = 0          # every DNS response received, any rcode (+ dns_rcode_<NAME>)
     dns_error: int = 0             # DNS exceptions that were not timeouts (types in the log)
 
@@ -317,6 +338,9 @@ class Orchestrator:
 
         # Latency accumulators (DORA + renew SEPARATELY, DNS resolve, both prop legs).
         self.lat_dora = LatencyAccumulator("dhcp_dora_ack")
+        # Late DORA ACKs (after the device gave up) get their own histogram so
+        # the strict one keeps its meaning and the slow tail is still visible.
+        self.lat_dora_late = LatencyAccumulator("dhcp_dora_ack_late")
         self.lat_renew = LatencyAccumulator("dhcp_renew_ack")
         self.lat_dns = LatencyAccumulator("dns_resolve")
         self.lat_prop_ipam = LatencyAccumulator("propagation_lease_to_ipam")
@@ -325,8 +349,10 @@ class Orchestrator:
         self.counters = Counters()
         self.devices: dict[int, Device] = {}
         self.online_set: set[int] = set()
-        # pending DHCP exchanges keyed by xid -> device index (recv correlation)
-        self.pending: dict[int, int] = {}
+        # Every in-flight (and recently closed) DHCP exchange keyed by xid: the
+        # reply is attributed to the exchange it answers, not to whatever the
+        # device happens to be doing when it arrives (#1057).
+        self.ledger = ExchangeLedger()
         self.dns_tally = DnsTally()
         # timer wheel: due_time -> list[(index, action)]
         self._timers: list[tuple[float, int, str]] = []  # min-heap of (when, index, action)
@@ -408,8 +434,15 @@ class Orchestrator:
     def _new_xid(self, dev: Device) -> int:
         # xid encodes the device index in the low 24 bits + a per-attempt nonce in the
         # high byte → cheap recv correlation while staying unique across re-sends.
-        nonce = self.rng.randrange(256)
-        return ((nonce & 0xFF) << 24) | (dev.index & 0xFFFFFF)
+        # The nonce must not collide with an exchange of this device the ledger
+        # still knows, or the reply would be attributed to the older send.
+        xid = dev.index & 0xFFFFFF
+        for _ in range(8):
+            nonce = self.rng.randrange(256)
+            xid = ((nonce & 0xFF) << 24) | (dev.index & 0xFFFFFF)
+            if self.ledger.get(xid) is None:
+                break
+        return xid
 
     def _send(self, pkt: bytes, dev: Device) -> None:
         sock = self._sock_for(dev)
@@ -429,7 +462,10 @@ class Orchestrator:
             self.log.debug("send failed: %s", exc,
                            extra={"fields": {"event": "send_error", "index": dev.index}})
 
-    def _send_discover(self, dev: Device) -> None:
+    def _send_discover(self, dev: Device, *, new_round: bool) -> None:
+        """A DISCOVER. ``new_round`` opens a fresh DORA round (arrival, re-arrival,
+        NAK); a retransmit after ``dora_timeout`` keeps the round, so a reply to
+        any of its sends still belongs to it."""
         if dev.discover_tpl is None:
             dev.discover_tpl = dp.build_discover(
                 mac=dev.mac, client_id=dev.client_id_bytes,
@@ -439,12 +475,18 @@ class Orchestrator:
         dev.xid = self._new_xid(dev)
         giaddr = self.subnets[dev.subnet_idx].giaddr if self.relay else None
         dp.patch_send_fields(pkt, xid=dev.xid, giaddr=giaddr)
+        if new_round:
+            dev.episode += 1
+            dev.lapsed = False
         dev.state = DState.DISCOVERING
         dev.tx_at = time.monotonic()
-        self.pending[dev.xid] = dev.index
+        self.ledger.open(dev.xid, dev.index, KIND_DISCOVER, dev.tx_at, dev.episode)
         self._send(bytes(pkt), dev)
         self.counters.dora_sent += 1
-        self._schedule(DORA_TIMEOUT_S, dev.index, "dora_timeout")
+        # The timer belongs to THIS exchange (#1057): it retransmits only if this
+        # send is still the device's live one when it fires — a DISCOVER's
+        # deadline no longer fires over the REQUEST that answered its OFFER.
+        self._schedule(DORA_TIMEOUT_S, dev.index, f"dora_timeout:{dev.xid}")
 
     def _send_request_renew(self, dev: Device) -> None:
         assert dev.leased_ip
@@ -454,11 +496,12 @@ class Orchestrator:
         dev.xid = self._new_xid(dev)
         dp.patch_send_fields(pkt, xid=dev.xid)  # NO giaddr — renew is unicast direct
         dev.state = DState.RENEWING
+        dev.episode += 1
         dev.tx_at = time.monotonic()
-        self.pending[dev.xid] = dev.index
+        self.ledger.open(dev.xid, dev.index, KIND_RENEW, dev.tx_at, dev.episode)
         self._send(bytes(pkt), dev)
         self.counters.renew_sent += 1
-        self._schedule(RENEW_TIMEOUT_S, dev.index, "renew_timeout")
+        self._schedule(RENEW_TIMEOUT_S, dev.index, f"renew_timeout:{dev.xid}")
 
     def _send_request_rebind(self, dev: Device) -> None:
         assert dev.leased_ip
@@ -470,10 +513,10 @@ class Orchestrator:
         dp.patch_send_fields(pkt, xid=dev.xid, giaddr=giaddr)
         dev.state = DState.REBINDING
         dev.tx_at = time.monotonic()
-        self.pending[dev.xid] = dev.index
+        self.ledger.open(dev.xid, dev.index, KIND_REBIND, dev.tx_at, dev.episode)
         self._send(bytes(pkt), dev)
         self.counters.rebind_sent += 1
-        self._schedule(RENEW_TIMEOUT_S, dev.index, "rebind_timeout")
+        self._schedule(RENEW_TIMEOUT_S, dev.index, f"rebind_timeout:{dev.xid}")
 
     def _send_request_selecting(self, dev: Device, offered_ip: str, server_id: str) -> None:
         pkt = bytearray(dp.build_request_selecting(
@@ -482,10 +525,18 @@ class Orchestrator:
         dev.xid = self._new_xid(dev)
         giaddr = self.subnets[dev.subnet_idx].giaddr if self.relay else None
         dp.patch_send_fields(pkt, xid=dev.xid, giaddr=giaddr)
-        dev.tx_at = time.monotonic()  # keep DORA clock running through SELECTING
-        self.pending[dev.xid] = dev.index
+        dev.tx_at = time.monotonic()  # the DORA latency is the REQUEST->ACK leg (unchanged)
+        self.ledger.open(dev.xid, dev.index, KIND_SELECT, dev.tx_at, dev.episode)
         self._send(bytes(pkt), dev)
-        self._schedule(DORA_TIMEOUT_S, dev.index, "dora_timeout")
+        self.counters.request_sent += 1
+        self._schedule(DORA_TIMEOUT_S, dev.index, f"dora_timeout:{dev.xid}")
+
+    def _schedule_t1(self, dev: Device) -> None:
+        """(Re)arm the T1 renewal for the lease the device holds now. The token
+        retires any earlier T1 timer, so a lease re-ACKed twice (a late renew
+        ACK and then the rebind's) renews once, not twice."""
+        dev.t1_token += 1
+        self._schedule(T1_RENEW_S, dev.index, f"t1_renew:{dev.t1_token}")
 
     def _send_release(self, dev: Device) -> None:
         if not (dev.leased_ip and dev.server_id):
@@ -499,12 +550,24 @@ class Orchestrator:
         self.counters.releases += 1
 
     # ---------------- FSM event handlers ----------------
-    def _on_offer(self, dev: Device, reply: dict) -> None:
+    def _on_offer(self, dev: Device, reply: dict, ex: Exchange | None, now: float) -> None:
+        self.counters.dora_offer += 1
+        if dev.state is not DState.DISCOVERING:
+            # A late or duplicate OFFER: the device already took a lease or gave
+            # up, so nobody is selecting. Counted so kea's OFFER tally reconciles
+            # with ours; the FSM does not move.
+            self.counters.offer_ignored += 1
+            if ex is not None and ex.open:
+                self.ledger.close(ex.xid, now, "offer")
+            return
         # SELECTING: accept the OFFER with a REQUEST(opt-50/opt-54).
         offered = reply.get("yiaddr")
         sid = reply.get("server_id") or offered
         if not offered or offered == "0.0.0.0":
             return
+        if ex is not None and ex.open:
+            # The DISCOVER is answered: its own timer must not retransmit it.
+            self.ledger.close(ex.xid, now, "offer")
         self._send_request_selecting(dev, offered, sid)
 
     def _ip_in_seeded_subnet(self, ip: str) -> bool:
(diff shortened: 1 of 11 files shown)


perf/generators/orchestrator/tests/test_device_fleet_accounting.py
```diff
@@ -0,0 +1,295 @@
...
+import argparse
+import asyncio
+import json
+import sys
+import types
+from pathlib import Path
+
+import pytest
+
+yaml = pytest.importorskip("yaml")
```
mzac, line 21: This is my main point on the PR: **the two test files that pin what this change fixes do not run in CI.**

`.github/workflows/ci.yml:765` installs pytest only, and `Makefile:479` mirrors it. Measured on this head with pytest alone:

```
56 passed, 2 skipped
SKIPPED [1] .../test_device_fleet_accounting.py:20: could not import 'yaml'
SKIPPED [1] .../test_collect_generator_rows.py:14: could not import 'yaml'
```

Those two are 404 of the ~550 new test lines, and they are the ones covering the *integration* where both #1057 bugs actually lived — `_on_ack` reaching the ledger, `_dns_query` counting rcodes, collect.py wiring the fold into b6/b6a/b7. The two that do run (`test_accounting.py`, `test_generator_tallies.py`, 15 tests) exercise the new pure modules in isolation and cannot see a regression in either changed file.

The negative control makes it concrete. Reverting **only** `device_fleet.py` and `collect.py` to `origin/main`, keeping the new modules and tests:

```
2 skipped in 0.01s      # the two behavioural files
15 passed in 0.01s      # the two pure files, unaffected
```

Not one failure. So both bugs could be reintroduced with the job green. Base is `41 passed, 0 skipped`, so this PR also introduces the first skipped tests in `perf/` — and the job's own comment says it exists because "#955 merged on 25 green checks that opened none of its files", which is this shape one level in.

The docstring's rationale ("CI's perf job installs pytest only — the pure tests carry the contract") is an accurate description of today, but the pure tests demonstrably do not carry *this* contract. Adding the deps to ci.yml + the Makefile costs 0.43 s of runtime.

perf/generators/orchestrator/tests/test_device_fleet_accounting.py
```diff
@@ -0,0 +1,295 @@
...
+    o.counters.dns_sent += 1
+    o.counters.dns_answered += 1
+    o._finalize()
+    path = o.rp.generator("orchestrator.shard0.summary.ndjson")
+    summary = json.loads(path.read_text().splitlines()[-1])
+    assert summary["counters"]["dora_ack"] == 1 and summary["counters"]["dns_rcode_REFUSED"] == 1
+    assert summary["handshake"]["strict_pct"] == 100.0 and summary["handshake"]["acked_late"] == 0
+    assert summary["dns"]["rcodes"] == {"REFUSED": 1} and summary["dns"]["ok"] == 0
+    assert summary["dora_ack_late"]["count"] == 0 and summary["dora_ack"]["count"] == 1
+    assert o.rp.generator("orchestrator.shard0.dhcp_dora_ack_late.hdr").exists()
```
mzac: Worth fixing together with the skip above, because it is the trap laid for whoever does it: this assertion needs **hdrhistogram**, which is not in the file's `importorskip` set.

Measured on this head:

- `pytest + pyyaml + dnspython` → `67 passed, 1 failed` — this line, `assert ... dhcp_dora_ack_late.hdr").exists()` is False.
- `+ hdrhistogram` → `68 passed`, reproducing your stated count exactly.

So the guard at the top of the file (yaml, dns.exception, dns.rcode) is one short of what the module asserts, and a partially-provisioned environment goes **red** rather than skipping. Concretely: someone reads the docstring's "needs PyYAML / dnspython", adds those two to the CI job to un-skip these tests, and turns the job red.

Either `pytest.importorskip("hdrhistogram")` beside the other three, or gate this one assertion on the histogram backend being present.

perf/harness/spddi_perf/collect.py
```diff
@@ -675,38 +676,64 @@ def criterion_b(rd: RunData) -> list[dict[str, Any]]:
...
+    # The orchestrator's REFUSED count joins dnsperf's: a refused query measured
+    # nothing, whether the name escaped the validator or the generator queried
+    # from a vantage outside the served view — the run cannot stand either way.
+    refused_dnsperf = sum(int((r.get("rcode") or {}).get("REFUSED", r.get("refused", 0)) or 0)
+                          for r in rd.dnsperf)
+    refused_orch = int(orch_rc.get("REFUSED", 0))
+    refused_total = refused_dnsperf + refused_orch
+    has_refused_data = bool(rd.dnsperf) or has_orch
     rows.append(_row(
         "b6a", "DNS REFUSED rate exactly 0 (structural — out-of-zone leak, §4.9)",
```
mzac, line 712: Flagging the consequence rather than the code, because I think the new verdict is correct: **this makes the perf gate FAIL on the current rig the moment it merges.**

`collect.py:1304` is `structural_ids = {"a1", "a4", "a6", "b6a", "d1", "d5"}` and :1308 returns `FAIL, f"structural invariant failed: {...}"` — so a b6a FAIL forces the overall verdict. With `has_refused_data = bool(rd.dnsperf) or has_orch`, any orchestrator run with `dns_rcode_REFUSED > 0` now fails structurally where it previously read NO_DATA. Your own live run is `606,237 REFUSED` → `overall FAIL — structural invariant failed: b6a`.

The diagnosis in the PR is that the generator queries from a vantage outside the served view — a property of the rig, which this change deliberately does not fix. And "Verification still to come" says the fleet-side proof is the pin bump "after which the cluster load tier is re-measured", so the first post-merge measurement is red by construction.

I'd merge this verdict change — a wholly-refused stream did measure nothing, and saying so is the point of the PR. But flipping a gate from NO_DATA to FAIL is a decision rather than a bug fix, so it would be easier to approve if the PR said whether the vantage fix is queued and what the gate is expected to report until it lands.

perf/harness/spddi_perf/collect.py
```diff
@@ -675,38 +676,64 @@ def criterion_b(rd: RunData) -> list[dict[str, Any]]:
...
+    # accounting its answers were invisible to b6/b6a: a run whose 606k queries
+    # BIND answered REFUSED read "ok 0, timeouts 46" and b6a said NO_DATA.
+    orch_c = gt.sum_counters(rd.orchestrator_summary)
+    orch_rc = gt.dns_rcodes(orch_c)
+    orch_answered = int(orch_c.get("dns_answered") or 0) or sum(orch_rc.values())
+    has_orch = bool(rd.orchestrator_summary)
+
+    # b6 — DNS SERVFAIL rate < 0.1% steady. dnsperf rcode + orchestrator rcode.
+    dp_sent, dp_servfail = _dnsperf_rcode_totals(rd, "SERVFAIL")
+    sf_total = dp_sent + orch_answered
```
mzac: Minor: this adds dnsperf's **sent** to the orchestrator's **answered**, so the rate's denominator is two different quantities.

`_dnsperf_rcode_totals` returns dnsperf's `sent`, while `orch_answered` is `dns_answered` — responses received, excluding timeouts and errors. On a run where the orchestrator times out heavily the denominator is smaller than the queries issued, so the printed SERVFAIL rate is biased high against the 0.1 % threshold.

The bias is in the conservative direction and the row's formula string states exactly what it does, so this is disclosable rather than wrong. But `dns_sent` is in the same counter set and would make both halves the same quantity — worth a deliberate call either way with a one-line comment saying which, since someone reconciling b6 against BIND's own ratio will otherwise not reproduce it.

mzac (commented):
Reviewed at `bacffb9` in a clean worktree. The accounting design is good and the ledger is the right primitive; most of my time went on the parts that carry the correctness claims, and they hold. Reporting what I checked so nobody re-derives it:

- **The verdict buckets are mutually exclusive** — `duplicate` returns, then `late` returns, then `dora_ack`, with `over_budget` as a *subset* marker rather than a fourth bucket. The exactness identity in your table (3,962 = 1,160 + 241 + 2,561) holds, and `give_up()` setting `expired_at` alongside `gave_up_at` is safe because `_on_dora_ack` tests `late` first.
- **The ledger is bounded** — `purge(now)` runs every scheduler tick, every close / expire / give-up path schedules one, and the stale-heap-entry handling is correct, including the case where `open()` supersedes a same-xid entry and the old heap entry resolves to the new exchange.
- **xid** (8-bit nonce << 24 | 24-bit index, probed 8 times, superseding on collision) keeps collisions intra-device and handles them.
- `with_late_pct` is clamped with `min(attempts, acked + late)`, so it cannot print >100 % even if the documented `acked_late <= timeouts` invariant were ever violated.
- `dns_timeouts_from_windows` takes the max per shard — correct for a cumulative counter and robust to out-of-order rows.
- **`MAX_DORA_RETRIES` still bounds the stated 2× change**: there is no separate SELECTING state, so a SELECT timeout lands in the same `dora_timeout` handler and increments the same counter.
- **No CHANGELOG needed** — I checked rather than assuming: 4 of the last 4 perf-only commits touch none.

@amoona6 — four points, all four inline. #1 is the one I'd want addressed here; #2 is the trap waiting for whoever fixes #1; #3 is a decision rather than a defect and may just need a sentence in the PR.

### Where the fix for #1 goes (no line in this diff to hang it on)

`.github/workflows/ci.yml:765` is `run: pip install pytest`, and `Makefile:479` mirrors it (`pip -q install pytest`). Adding `pyyaml dnspython hdrhistogram` to both un-skips the two files; measured runtime with all of them installed is **0.43 s**, so nothing is bought by leaving them out.

---

Mechanically: 19 green + 1 skipped, `MERGEABLE`, `BEHIND` main.

perf/generators/orchestrator/tests/test_device_fleet_accounting.py
amoona6: Confirmed and reproduced (pytest alone: 56 passed / 2 skipped; your negative control with the deps installed: 12 failures, with pytest alone: 0). Addressed in bb8074a7: ci.yml installs pytest, PyYAML, dnspython and hdrhistogram; the Makefile mirror does the same, taking hdrhistogram as a binary wheel when one exists (its sdist needs a C toolchain that python:3.12-slim lacks on arm64). After: 69 passed, 0 skipped in the ci.yml shape. Details in the summary comment.

perf/generators/orchestrator/tests/test_device_fleet_accounting.py
amoona6: Confirmed (pyyaml + dnspython without hdrhistogram: 67 passed / 1 failed at this line). Addressed in bb8074a7 by gating the assertion on the histogram backend rather than the import — `assert hdr.exists() == lifecycle_log._HAVE_HDR` — so a partially provisioned environment passes (measured: 69 passed without hdrhistogram) instead of going red; the importorskip set stays yaml + dns, the two hard requirements.

perf/harness/spddi_perf/collect.py
amoona6: Agreed that it is a decision, and it is now stated in the PR body (section "What the gate reports until the vantage is fixed"): the vantage fix is queued on the QA side (ddi-pg #216, an in-view source address for the generator's DNS stream); until it lands every out-of-view run reads `overall FAIL — structural invariant failed: b6a` by construction, which is the honest reading (the run measured nothing), and the QA harness turns its DNS rows into an infrastructure skip naming the vantage (ddi-pg #209, merged) and will treat the SLO row the same way (ddi-pg #219). Once in view, b6a measures the out-of-zone leak it was written for. No code change here.

perf/harness/spddi_perf/collect.py
amoona6: Confirmed. Addressed in 475d53c3: both halves are rated over queries sent (`dns_sent`), with a comment saying why — a query that timed out was sent and did not SERVFAIL, so the rate is per attempt on either side and reproduces BIND's own SERVFAIL / QUERY ratio. A test pins it (8 SERVFAILs in 10,000 sent of which 8,000 answered: 0.08 % PASS where the answered denominator read 0.10 % FAIL). Our four live runs had SERVFAIL 0, so their rows read 0 % either way.

perf/generators/orchestrator/tests/test_accounting.py
```diff
@@ -0,0 +1,146 @@
...
+    c = led.close(xid(2), 9.0, "ack")                      # kea's ACK, 5 s after the REQUEST
+    assert c is not None and c.late and c.over_budget and not c.duplicate
+    assert c.exchange.gave_up_at == 8.1
+    again = led.close(xid(2), 9.5, "ack")                  # a retransmitted ACK
+    assert again is not None and again.duplicate
+
+
+def test_unknown_xid_is_none_and_purge_forgets_after_grace() -> None:
+    led = ExchangeLedger(grace_s=10.0)
+    assert led.close(xid(9), 1.0) is None
```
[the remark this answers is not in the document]
amoona6: Fixed in 4a57c9e4: the call is made into a local first and the assert reads the local.

perf/generators/orchestrator/tests/test_accounting.py
```diff
@@ -0,0 +1,146 @@
...
+    led.open(xid(1), DEV, KIND_DISCOVER, 0.0)
+    led.close(xid(1), 1.0, "nak")
+    assert led.purge(5.0) == 0 and len(led) == 1           # still known within the grace
+    assert led.purge(11.5) == 1 and len(led) == 0
+    assert led.close(xid(1), 12.0) is None                 # now unmatched
```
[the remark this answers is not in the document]
amoona6: Fixed in 4a57c9e4: the call is made into a local first and the assert reads the local.

perf/generators/orchestrator/device_fleet.py
```diff
@@ -66,18 +66,30 @@
...
     import dns.rcode as _dns_rcode  # type: ignore
 
     _HAVE_DNSPYTHON = True
+    # A timed-out query is a timeout; every other exception is an error the
+    # tally names by type (#1057 — the two used to share one counter).
+    _DNS_TIMEOUT_EXC: tuple[type[BaseException], ...] = (
+        _dns_exc.Timeout, asyncio.TimeoutError, TimeoutError)
 except Exception:  # pragma: no cover - exercised only in a bare env
-    _dns_aq = _dns_msg = _dns_rcode = None  # type: ignore
+    _dns_aq = _dns_exc = _dns_msg = _dns_rcode = None  # type: ignore
```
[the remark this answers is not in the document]
amoona6: Fixed in 4a57c9e4: the binding is gone; the timeout class is imported as `_DnsTimeout` and used in the exception tuple, and the bare-environment branch binds nothing it does not read.

